PR #5, opened 2019-08-17
docs(readme): first draft for workflow documentation

mostaphaRoudsari:
@AntoineDao and @chriswmackey, While I'm working on the generator and validator code I thought I should share the documentation with you first and get some feedback. See the README.md.

@chriswmackey I know you haven't been involved in the process of coming up with this workflows much but this will actually make you a great candidate to give us some feedback. This workflow structure will be used for creating workflows for all the different packages.

PS: apparently this approach has a name and is called `Documentation Driven Development`!

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,11 +68,11 @@ value for inputs or it can be passed to workflow using an input YAML / JSON file
       value: /mnt/c/ladybug/untitled
     artifacts:
     - name: sky_file
-      value: full_path_to_sky_file.sky
+      path: full_path_to_sky_file.sky
     - name: material_file
-      value: full_path_to_material_file.mat
+      path: full_path_to_material_file.mat
     - name: geometry_file
-      value: full_path_to_geometry_file.rad
+      path: full_path_to_geometry_file.rad
 
 ```
 
@@ -82,27 +82,16 @@ value for inputs or it can be passed to workflow using an input YAML / JSON file
 image. This is an operator for running arbitrary Radiance commands.
 
 ```yaml
-
-operators:
   - name: radiance-operator
-    inputs:
-      parameters:
-        - name: command
     container:
       name: radiance52
       image: ladybugtools/radiance:5.2
-      command:
-        - /bin/sh
-        - -c
-      args:
-        - '{{ inputs.parameters.command }}'
-    requirements:
+    local:
       app:
         - name: radiance
           version: ">=5.2"
           command: rtrace -version  # command to get the version
           pattern: r'\d+\.\d+'  # regex pattern to extract the version from command output
-
 ```
 
 Here is another one for running a `template` for running `ladybug` Python scripts.
@@ -112,17 +101,10 @@ by removing it from the list.
 ```yaml
 
   - name: ladybug-operator
-    inputs:
-      parameters:
-        - name: script
     container:
       name: ladybug
       image: ladybugtools/ladybug:latest
-      command:
-        - python
-      args:
-        - '{{ inputs.parameters.script }}'
-    requirements:
+    local:
       pip:
         - name: lbt-ladybug
           version: ">=0.4.3"
@@ -137,14 +119,12 @@ by removing it from the list.
 ```
 
 You can also import the operator from a `YAML` or `JSON` file using the `import_from`
-key.
+key. All the fields except for the name will be loaded from the file. It's good practice
+to use the same name inside the yaml file and the workflow.
 
 ```yaml
 
   - name: ladybug-operator
-    inputs:
-      parameters:
-        - name: script
     import_from: /operators/ladybug_operator.yaml
 
 ```
@@ -164,6 +144,52 @@ Templates can use the `import_from` keyword to import the `template` from a `YAM
 `JSON` file. Keep in mind that a template in a file should be self-sufficient and include
 all the information including `operator`s. This is different from when the `template` is
 part of the `workflow` where an `operator` will be referenced by name.
+
+Here is a `module` for generating a sky with desired irradiance which is part of a
+workflow. In order to make it a valid `template` in a separate file you should use the
+full operator specification from the last step instead of only using the name of the
+operator.
+
+```yaml
+  name: generate_sky
+  inputs:
+    parameters:
+      - name: desired_irradiance
+        value: 100000   # default value which can be overwritten
+  operator: radiance-operator
+  # commands and args will be used both locally and inside the container
+  command:
+    - /bin/sh
+    - -c
+  args:
+    - gensky -c -B '{{ inputs.parameters.desired_irradiance }}' > '{{ workflow.artifacts.sky }}'
+  outputs:
+    artifacts:
+      - name: sky
+        path: '{{ workflow.artifacts.sky }}'
+```
+
+Alternatively, you can use `honeybee-radiance` operator which has a command for
+generating sky.
+
+```yaml
+  name: generate_sky
+  inputs:
+    parameters:
+      - name: desired_irradiance
+        value: 100000   # default value which can be overwritten
+  operator: honeybee-radiance
+  command:
+    - honeybee-radiance
+  args:
+    - fixed-irradiance-sky
+    - --irradiance
+    - "{{inputs.parameters.desired_irradiance}}"
+  outputs:
+    artifacts:
+      - name: sky
+        path: '{{ workflow.artifacts.sky }}'
+```
 
 ## 4. flow
 
@@ -192,7 +218,7 @@ flow:
       template: generate_sky
       inputs:
         parameters:
-          - name: desired_illuminance
+          - name: desired_irradiance
             value: 100000
     - name: generate_octree_step
       template: generate_octree
@@ -244,7 +270,7 @@ flow:
       template: generate_sky
       inputs:
         parameters:
-          - name: desired_illuminance
+          - name: desired_irradiance
             value: 100000
     - name: generate_octree_task
       template: generate_octree
--- a/README.md
+++ b/README.md
@@ -309,3 +309,51 @@ flow:
 Several files might be generated in the `flow` section and outputs indicates which ones
 should be considered as the final outputs of the workflow. Executors may return back
 these outputs as a dictionary.
+
+
+# Command Line Interface
+
+You can also use queenbee from command line. The most commonly used commands are:
+
+1. validate
+  
+  `queenbee validate [WORKFLOW-FILE]`
+
+  This command validates the workflow to ensure:
+    a. the workflow complies with queenbee schema
+    b. all the `import_from` resources are available and valid.
+    c. all the operators are included in workflow file.
+
+  You can also validate a workflow against an input file.
+
+  `queenbee validate [WORKFLOW-FILE] --inputs [INPUT-FILE]`
+
+2. package
+  
+  `queenbee package [WORKFLOW-FILE] [PACKAGED-WORKFLOW-FILE]`
+
+  This command packages the workflow and all its dependencies into a single file. If
+  there is no `import_from` in original workflow the packaged workflow will be identical
+  to the original workflow.
+
+3. freeze
+  
+  `queenbee freeze [WORKFLOW-FILE] [INPUT-FILE] [FROZEN-WORKFLOW-FILE]`
+
+  Queenbee workflows are designed to be reusable and it is valid to have input parameters
+  with no default values. These values will be provided in an input file. In some cases
+  you want to use the workflow over and over with the same input values. This command
+  makes a frozen version of the workflow with all the values hard-coded in the workflow.
+
+  `freeze` command calls `package` command under the hood. In other words a frozen
+  workflow will not have an `import_from` key.
+
+4. populate-inputs
+
+  `queenbee populate-inputs [WORKFLOW-FILE] [INPUT-FILE]`
+
+  This command generates a template input file for a specific workflow. You can use
+  `--include-defaults` flag to also get the inputs that already have a default value. The
+  inputs with referenced inputs with prefix variable name (e.g.
+  {{steps.create_octree.outputs.artifacts.octree}}) will not be included in the inputs
+  file. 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,2 +1,285 @@
...
+        - name: command
+    container:
+      name: radiance52
+      image: ladybugtools/radiance:5.2
+      command:
+        - /bin/sh
+        - -c
+      args:
+        - '{{ inputs.parameters.command }}'
+    requirements:
```
AntoineDao: Does this requirements section check for specific versions in the current system or does it also try to install it for you?

README.md
```diff
@@ -1,2 +1,285 @@
...
+image. This is an operator for running arbitrary Radiance commands.
+
+```yaml
+
+operators:
+  - name: radiance-operator
+    inputs:
+      parameters:
+        - name: command
+    container:
```
AntoineDao: Is this a mandatory field?

README.md
```diff
@@ -1,2 +1,285 @@
...
+    import_from: /operators/ladybug_operator.yaml
+
+```
+
+## 3. templates
+
+Templates are a discrete reusable unit of code that can be executed separately or as part
+of a workflow. Queenbee supports 4 type of objects as templates:
+
+- `module` : A single module. Each module uses an operator from operators section.
```
AntoineDao: Is a module essentially 1 workflow task such as `create skyfile` or `run gridbased simulation`?

README.md
```diff
@@ -1,2 +1,285 @@
...
+      template: generate_sky
+      inputs:
+        parameters:
+          - name: desired_illuminance
+            value: 100000
+    - name: generate_octree_step
+      template: generate_octree
+      inputs:
+        parameters:
+           - name: scene_files
```
AntoineDao: How do we differentiate between parameters that are arrays and parameters that are single line strings? Can a parameter also be a JSON?

README.md
```diff
@@ -1,2 +1,285 @@
...
+
+```yaml
+
+  inputs:
+    parameters:
+    # this is just the path to target folder not the content itself. It can be referenced
+    # as {{workflow.parameters.folder}}
+    - name: folder
+      value: /mnt/c/ladybug/untitled
+    artifacts:
```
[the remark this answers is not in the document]
mostaphaRoudsari: They can be urls. There is a boolean to indicate if an artifact is a url. Files are actually interchangeable between `parameters` and `artifacts` for local runs. The distinction will be important when converted to an Argo workflow.

```python
class Artifact(BaseModel):
    """Artifact indicates an artifact to place at a specified path"""

    name: str = Schema(
        ...,
        description='name of the artifact. must be unique within a task\'s '
        'inputs / outputs.'
    )

    path: str = Schema(
        None,
        description='Path is the path to the artifact which can be a local path or a url.'
    )

    is_url: bool = Schema(
        False,
        description='Switch to indicate if path is a url.'
    )
```

README.md
mostaphaRoudsari: This is out of the scope of `queenbee` and should be decided when one runs the study. I have been thinking it should be an option in the command.

`workerbee [WORKFLOW] [INPUT_FILE] --install-requirements`

README.md
mostaphaRoudsari: no. only name is at this point which means you can create an operator that only runs locally or only runs using a container. Maybe I should break down operators to be two different types for `LocalOperator` and `DockerOperator`?

README.md
mostaphaRoudsari: Yes. `module` is the smallest executable unit of templates. I used to call them `template` but it was confusing as all of these different types are supposed to be used as templates.

README.md
mostaphaRoudsari: Yes. It can be from any types including JSON. The template should take care of handling different types.

```python
class Parameter(BaseModel):
    """Parameter.

    Parameter indicate a passed string parameter to a service template with an optional
    default value.
    """
    name: str = Schema(
        ...,
        description='Name is the parameter name.'
    )

    value: Any = Schema(
        None,
        description='Default value to use for an input parameter if a value was not'
        ' supplied.'
    )
```

At one point I considered handling them in workflow by adding a `*List` type but then I felt it is out of the scope of a task / step and should be handled in each template itself.

```python
class ArtifactList(Artifact):
    """Artifact List indicates a list of artifact locations."""
    path: List[str] = Schema(
        None,
        description='Path is the path to the artifact which can be a local path or a url.'
    )

    join: str = Schema(
        None,
        description='A string to join inputs together into a single string.'
    )
```

README.md
```diff
@@ -1,2 +1,285 @@
...
+## 2. operators
+
+`Operators` include the requirements for running `templates` locally or using a Docker
+image. This is an operator for running arbitrary Radiance commands.
+
+```yaml
+
+operators:
+  - name: radiance-operator
+    inputs:
```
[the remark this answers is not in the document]
mostaphaRoudsari: Good point! I checked the operator again and actually the schema doesn't have an input. The reason for adding it here, I think, was that I needed to provide a solution to parametrize the command inside the container.

Let me think about it. It is also possible to remove them and argue that all the parameters in container should fall back to module inputs and module command or even argue that command in container should be the module command and one should take care of all the parameters in the module. This is actually how the implementation is right now.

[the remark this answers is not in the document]
mostaphaRoudsari: Yes. I think that is how it should be. I should remove the inputs from here.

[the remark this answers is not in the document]
mostaphaRoudsari: I tested a couple of options and looks like the most elegant one is to take the operator back under `module` template itself. The reason for having them all in one place was to:

1. avoid duplication
2. make it easier to validate the requirements before runs (basically check it once in one place)

But now I can see that this idea introduces limitations for containers as now all of the templates should comply with the exact same `args` and `command` input.

To address duplication one can use `import_from` key to import them from a file if the container / requirements is exactly the same for different modules.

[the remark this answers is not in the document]
mostaphaRoudsari: See the updated version! It should address both concerns there and keeps `command` and `args` only inside the `module`.

README.md
AntoineDao: My main reason for bringing this up is because it seems very ambitious to include automated install of arbitrary software.

README.md
AntoineDao: Nah i think it's fine to keep them in as optional rather than two different types of operators.

README.md
AntoineDao: Have you tried this with Argo? I'm not sure argo can take non string inputs.

README.md
mostaphaRoudsari: Thanks. This is how it ended up being. see the updated documentation.

README.md
mostaphaRoudsari: I have not. based on the examples it should be able to handle other types of values too. I will double check. If that was the case then we have to use the `join` option to join them together as a string but I think it should not be necessary.

README.md
mostaphaRoudsari: added examples for modules to the document.

README.md
mostaphaRoudsari: Good point probably we should add a section for user to provide installation commands or otherwise throw an exception. Added an issue to address this in the future.

README.md
```diff
@@ -1,2 +1,359 @@
...
+Alternatively, you can use `honeybee-radiance` operator which has a command for
+generating sky.
+
+```yaml
+  name: generate_sky
+  inputs:
+    parameters:
+      - name: desired_irradiance
+        value: 100000   # default value which can be overwritten
+  operator: honeybee-radiance
```
chriswmackey: Where is this `honeybee-radiance` operator and what does it look like?  I can see what the `radiance-operator` looks like in the `operators` section.  At least give me a link to this `honeybee-radiance` operator so that I know what you are talking about here.

README.md
```diff
@@ -1,2 +1,359 @@
...
+```
+
+## 4. flow
+
+`flow` is the heart of the workflow to define how the templates should be executed.
+Queenbee supports two different types of workflows. DAGFlow and SteppedFlow. 
+
+It's common to use the output of one step as an input for another step or reference one
+of the workflow inputs as an input for one of the steps or tasks. Queenbee supports below
+words as prefix variable name:
```
chriswmackey: Should be:
`Queenbee supports the following words as prefix variable names:`

README.md
```diff
@@ -1,2 +1,359 @@
...
+    - name: sky_file
+      path: full_path_to_sky_file.sky
+    - name: material_file
+      path: full_path_to_material_file.mat
+    - name: geometry_file
+      path: full_path_to_geometry_file.rad
+
+```
+
+## 2. operators
```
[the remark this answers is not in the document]
chriswmackey: Or, if there is some reason for this order, it might be nice to just mention it in the docs here.

README.md
```diff
@@ -1,2 +1,359 @@
...
+words as prefix variable name:
+
+- workflow: "{{workflow.xx.yy}} is used for workflow level parameters.
+- steps: "{{steps.step_name.xx.yy}} is used in chained templates.
+- tasks: "{{tasks.task_name.xx.yy}} is used in DAG task to refer to other tasks.
+- inputs: "{{inputs.xx.yy}}" is used in operators.
+- item: "{{item}}" or "{{item.key_name}}" is used in loops. You can change item to a
+  different keyword by setting up `loop_var` in `loop_control`.
+
+Here is a example of creating the sky and generating and octree as two consecutive steps.
```
chriswmackey: Should be:
`Here is a example of creating a sky and generating an octree`

README.md
```diff
@@ -1,2 +1,359 @@
...
+           - name: scene_files
+             value:
+               - "{{steps.generate_sky_steps.outputs.sky_file}}"
+               - "{{workflow.inputs.parameters.folder}}"/model/static/scene.mat
+               - "{{workflow.inputs.parameters.folder}}"/model/static/scene.rad
+
+```
+
+We can add generating sensors and running the ray-tracing itself to the current flow but
+there is a difference between these new two steps and what we had before. In previous
```
chriswmackey: The grammar of these two sentences is making it tough for me to understand what you mean to say but I think you mean something like this:
`We can generate sensors and run the ray-tracing using the current flow but
there is a difference between these two new steps and what we had before`.
I have to be honest that I don't know what `we had before`.  Do you mean the example above?

README.md
```diff
@@ -1,2 +1,359 @@
...
+            value: "{{tasks.generate_octree_task.outputs.parameters.octree}}"
+      loop: "{{ tasks.generate_grids_task.outputs.parameters.grids }}"
+```
+
+## 5. outputs
+
+`outputs` is a list of parameters or artifacts that are generated during the flow.
+Several files might be generated in the `flow` section and outputs indicates which ones
+should be considered as the final outputs of the workflow. Executors may return back
+these outputs as a dictionary.
```
chriswmackey: I think you haven't explained this correctly because I am still confused if `outputs` just refers to the files that are generated during the flow or if `outputs` is a JSON /YAML (or a dictionary?) that notes the output files that you want to save.  I feel like it may be the latter so I would recommend including an example of what this looks like and changing the wording to something like this:

Several files might be generated in the `flow` section and the `outputs` section indicates which ones
should be saved as the final outputs of the workflow. Executors may return these output file paths in a dictionary.

README.md
```diff
@@ -1,2 +1,359 @@
...
+Here we discuss the generic structure of a Queenbee workflow. See OpenApi schema
+documentation for all Workflow objects: [Missing link]()
+
+```
+
+  Workflow
+    |
+    |__ inputs
+    |   |___ parameters
+    |   |___ artifacts
```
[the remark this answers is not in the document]
mostaphaRoudsari: parameters in this case are like arguments in a mathod. basically a collection of `key:value`. you can see it as `Workflow(parameters, artifacts)`.

`artifacts` are actually what can be called `assets`. I used artifacts in this case as this is what Argo uses and is the common word is pipelines and workflows. Look up for `S3 artifacts` as an example.

README.md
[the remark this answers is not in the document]
mostaphaRoudsari: I see your point and I agree. I will add a note to the docs to mention this. The reason is that naturally you want to have the objects that gets used / referenced by other objects higher in the file. So operators come first and then templates which reference these operators.

README.md
mostaphaRoudsari: There is none in this document. I can update the operator for ladybug to be honeybee-radiance operator! I didn't include it as it will be a long one but it makes sense to have one if we are going to reference it.

mostaphaRoudsari:
```yaml

  - name: honeybee-radiance
    container:
      name: honeybee-radiance
      image: ladybugtools/honeybee-radiance-workflow:latest
    local:
      app:
        - name: radiance
          version: ">=5.2"
          command: rtrace -version
          pattern: r'\d+\.\d+'
      pip:
        - name: lbt-honeybee
          version: ">=0.4.3"
        - name: honeybee-radiance-workflow
      language:
        - name: python
          version: ">=3.6"
      platform:
       - Windows
       - Linux
       - Mac

```

also added to the document. thanks.

README.md
mostaphaRoudsari: we had before = the first two steps that we had as our steps before.

mostaphaRoudsari: What about this:

Now let's think about a longer workflow which also includes ray-tracing using the
generated octree. We need to add two new steps to the workflow:

1. generate sensor grids
2. run ray-tracing

But there is a difference between these new two steps and the initial two steps.
In the first two steps the second step of generating the octree was dependant on the
first step and it couldn't be executed until generating sky is finished. However for
generating the sensor grids we we do not need to wait for generating sky to be finished.
Finally, the last step of ray-tracing will need both the grid and the octree.

README.md
AntoineDao: I suspect that outputs just like inputs can have a `parameters` and an `artifacts` keys to specify the type of output.

mostaphaRoudsari: Yes and yes, and your wording is much better. I'm going to use it.

I also initially thought of outputs just as files but it can actually be more than just files. For instance a LEED workflow might also return a `parameter` for `passed` which can be `true` or `false`.

mostaphaRoudsari:
@AntoineDao, @chriswmackey, thank you for your input. I addressed them as much I as I could for now. Since we have had the first round of the discussion here I'm going to merge this in. I will send new revisions as new PRs.

PR #8, opened 2019-10-19
Add Queenbee schema

### LINKED ISSUE #1: Support loops in Tasks
See here: https://github.com/ladybug-tools/workerbee-dag/pull/3#discussion_r307993937

### LINKED ISSUE #2: finish updates in workflow structure
workflow structure is now in the process of being finished. I'm opening these issues so we can keep track of the tasks that are needed to be done to get the workflow in place.

- [x] add workflow level inputs [ and maybe outputs?]. having outputs makes it possible to use a workflow as a template.
- [x] finalize input binding between inputs file and DAG tasks. finding a neat solution here is critical to make the workflows flexible and useful.
- [x] update workflow class based on `inputs`, `operators`, `templates`, `dag` and `outputs`. it is currently builds of DAG which is fine but does make from and to file process quite messy.

### LINKED ISSUE #3: add support for `{{some-list}}` input for inputs
This is required for mapping inputs from one task to another.

### LINKED ISSUE #4: add simple examples of workflows
Which I can also use for testing!

mostaphaRoudsari:
This PR adds all different schemas for Queenbee objects.

It also updates the ReadMe documentation to match the latest schema specifications.

The next step is to add a Python client side for people to create the workflows more naturally. This is what I have in mind.

```python
from queenbee import Workflow, Task
from queenbee.operator.lib import RadianceOperator, CLIOperator


wf = Workflow('my_workflow')

task_1 = Task('task_1', cmd='ls -a', operator=CLIOperator)
task_2 = Task('task_1', cmd='rm -r *', operator=CLIOperator)

task_1.run_after(task_2)

wf.add_tasks([task_1, task_2])

wf.write('workflow.yaml')

```

resolves #1 
resolves #2 
resolves #3 
resolves #4


dev-requirements.txt
```diff
@@ -0,0 +1,9 @@
+coverage==4.5.2
+coveralls==1.5.1
+pytest==4.1.0
+pytest-cov==2.6.1
+Sphinx==1.8.5
+sphinx-bootstrap-theme==0.6.5
+sphinxcontrib-fulltoc==1.2.0
+sphinxcontrib-websupport==1.1.0
+twine==1.13.0
```
AntoineDao: Mind adding this dependency also:

> attrs==19.1.0

I just ran the tests with a fresh virtual environment and it seems a dependency of pytest was updated early October and breaks things unless you explicitly add the previous version... :upside_down_face:

mostaphaRoudsari: Added! Thanks.

mostaphaRoudsari:
Hi @AntoineDao, Thank you for the review. Sorry for the tests 🙈! It should be fixed now.

> workflow input artifcats: are these intended to be used as a root path for individual task input/outputs? That it my understanding from the README, however I couldn't find an example of a task level input with an artifact root path. Same goes for ouputs. I'm interested in seeing an example as this could be quite a confusing feature depending on how it is explained/represented.

Did you see this sample workflow and the comments? Does it address your concerns. Let me know if you think I should add more explanation to ReadMe.

https://github.com/ladybug-tools/queenbee/pull/8/files#diff-933912753cf857b7e56eb63ca50f1f66R12-R19

> Operator names: I understand the need for a name for each operator however using a name for the container seems redundant

Good point! Image name should be enough. I can use the name of the operator when converting them to Argo. I changed this. 

> Operator local

I think about this more and comment after our call.

> Command Line Interface: It took me a solid 15 mins to realise command line wasn't working because the code hasn't been written yet 🙃 I think in future it might be worth documenting what is actually coded in a readme and updating it as features are delivered.

Sorry! This turned out to be the limitation of Document-Driven-Design. Added a TODO comment to ReadMe.

AntoineDao:
I reran through the example and this answers my question:

https://github.com/ladybug-tools/queenbee/pull/8/files#diff-04c6e90faac2675aa89e2176d2eec7d8R235-R240

We manually implement the templated value into the path of other inputs. That's cool and makes sense to me! :smile:

mostaphaRoudsari:
> Operator local: This is the trickiest part of the queenbee/workerbee app... managing to run commands locally but checking for versions, dependencies etc... is complex. The Readme did not explain what the difference between a pip, python and app local operator is. Futhermore, as I think of this, I reckon a local app should just be an entrypoint into a CLI command and that's it. You could potentially add a link to an install instruction or something.

Agreed! Operator local has been a challenge to figure out. I like your suggestion for instruction / docs. We can use them like an exception message but the entry-point idea may not play well with `command` input. It also won't play well with tools like Radiance. For instance is `rtrace` the entry-point or `radiance`? Also how does one checks version?

I can add more details about app (executable application) versus pip (Python package) versus language (programming language) versus npm (node package) but I'm not sure that will address your concerns.

Also I'm realizing that `language` and `app` are almost the same and can be merged as one. So we will end up with:

1. `app` for applications that must be installed before using the commands or it will raise an error
2. Python / Node packages which Workerbee will try to install before running the workflow.

In that case one can also disregard installation checks by passing `--no-local-check` flag.

What do you think?

queenbee/schema/function.py
```diff
@@ -0,0 +1,151 @@
...
+        input_names = [param.name for param in values['inputs'].parameters]
+        for v in output_params:
+            match = var_parser(v.path)
+            cls.validate_variable(match, func_name, input_names)
+
+        return v
+
+    @staticmethod
+    def validate_variable(variables, func_name, input_names):
+        """Validate referenced variables."""
```
chriswmackey: It might be good to list what validation actually entails in the docstring. Also it seems odd to have a validation function that is not a `@validator`. Might be a good idea to explain why this is the case in the docstring.

queenbee/schema/operator.py
```diff
@@ -0,0 +1,135 @@
...
+    )
+    
+    version: str = Schema(
+        None,
+        description='Language version requirements. For instance ==3.7 or >=3.6'
+    )
+    
+    _valid_languages: Set[str] = set(['python', 'nodejs', 'bash'])
+
+    # NOTE: yaml convertion doesn't play well with Enum. Use validators like this
```
chriswmackey: Should be `conversion`.

queenbee/schema/operator.py
```diff
@@ -0,0 +1,135 @@
...
+                v, '\n - '.join(cls._valid_languages)
+            )
+        return v.lower()
+
+
+# TODO: This will most likely change. Too much to take on!
+class App(BaseModel):
+    """Local application."""
+    name: str = Schema(..., description='App name')
+    version: str = Schema(
```
chriswmackey: PEP8 👮 says that you should have one blank line. Same goes for the `Package` class.

queenbee/schema/operator.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+
+    npm: List[Package] = Schema(
+        None,
+        description='List of npm packages that are required for operator to '
+            'execute the commands locally. You must follow npm install requirements: '
+            'https://docs.npmjs.com/cli/install#synopsis'
+        )
+
+    # NOTE: yaml convertion doesn't play well with Enum hence using a validator.
```
chriswmackey: `conversion`.
I know. English is weird.

queenbee/schema/workflow.py
```diff
@@ -0,0 +1,81 @@
...
+
+    outputs: Arguments
+
+    # @validator('flow')
+    # def check_templates(cls, v, values):
+    #     """Check templates in flow to ensure they exist."""
+    #     template_names = set(t.name for t in values['templates'])
+    #     for task in v.tasks:
+    #         if task.template not in template_names:
+    #             raise ValueError('{} is not a valid template.'.format(task.template))
```
chriswmackey: I'd say it's generally bad practice to leave commented code in a commit. If you want to keep it here for future reference, I'd at least add a `TODO:` at the top of it.

AntoineDao: :police_car: :policeman: :police_car:

AntoineDao:
I see where you're trying to get with all of these points. I guess this is worth a slightly more in depth conversation. I think we're in agreement that we want a workflow schema that allows people to reliably execute tasks on their local machine + cloud. However, I think that your ambitions for how this would work on a local machine are... well... too ambitious. 

I think the strong installation/version control features you want to set up will be your downfall in terms of work and have entirely been solved by Docker containers (which is an option as part of the workflow). Therefore I think we should drop these features entirely from the workflow schema in favor of containers. If people can't/don't want to use containers then they will be responsible for managing the CLI tools used by the workflow.

## Entry Point
The fact this won't work with a `radiance` operator is fine in my opinion. From there we have 2 options:
1. break it down into a `rtrace`, `rpict` etc... operators
2. set a blank entrypoint for the radiance operator

## Installation
I have resumed my thoughts on this above but to drill it down a bit further: I have no idea how you plan on managing installation of packages on multiple OSs (not even talking windows vs Mac but windows 8 vs windows 10 vs windows 10 with minimal admin privilege etc...). In my opinion this is an endless black hole of dev time that won't reap much benefit. However, if you manage to pull it off we should change business and compete directly with package managers like apt, choco, npm, pip, gem, brew etc...

## Version Check
I see the benefit of adding a version control check to operators, especially if we're not managing install. Is it possible to set this as an optional field? I can think of a few cases where version checking will not be available through command line (eg: radiance).

queenbee/schema/arguments.py
```diff
@@ -0,0 +1,89 @@
...
+
+class Parameter(BaseModel):
+    """Parameter.
+
+    Parameter indicate a passed string parameter to a service template with an optional
+    default value.
+    """
+    name: str = Schema(
+        ...,
+        description='Name is the parameter name.'
```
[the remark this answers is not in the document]
mostaphaRoudsari: I'm not sure. It actually depends on the context. Arguments object is used both for inputs and outputs for different objects so it is hard to come up with a global rule.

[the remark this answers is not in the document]
mostaphaRoudsari: Thinking about it more enforcing them to be unique won't hurt. I added that to description.

queenbee/schema/dag.py
```diff
@@ -0,0 +1,104 @@
...
+
+    name: str = Schema(
+        ...,
+        description='A unique name for this dag.'
+    )
+
+    target: str = Schema(
+        None,
+        description='Target are one or more names of targets to execute in a DAG. '
+            'Multiple targets can be specified as space delimited inputs.'
```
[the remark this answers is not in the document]
mostaphaRoudsari: This was a copy / paste from Argo's definition and I agree it is hard to understand. I updated it to this:

```python

    target: str = Schema(
        None,
        description='Target are one or more names of target tasks to execute in a DAG. '
            'Multiple targets can be specified as space delimited inputs. When a target '
            'is provided only a subset of tasks in DAG that are required to generate '
            'the target(s) will be executed.'
    )
```

queenbee/schema/function.py
mostaphaRoudsari: This is a helper method that is used by a number for `@validator` classmethods. That's why it is not created as a typical Pydantic validator. I added a better description to docstring.

queenbee/schema/operator.py
mostaphaRoudsari: Thanks!

queenbee/schema/operator.py
mostaphaRoudsari: I have made everyone a PEP8 👮‍♂️! So proud of myself for this one. 😂 Thanks.

queenbee/schema/workflow.py
mostaphaRoudsari: Added a TODO and totally agreed!

mostaphaRoudsari:
## local requirements + entry point + installation

I'm fine with considering the local field as an optional informational field where one can provide information on what needs to be installed or how to install them. No version check, etc. and user will be responsible for installing requirements if running the workflow locally.

## versioning

Good idea! Added.

PR #10, opened 2019-11-07
Artifact rejig

AntoineDao:
This PR contains a series of changes to how artifacts are managed in the Queenbee schema to allow for manipulation of non local filesystem artifacts (things such as S3 buckets for example).

The main changes to the schema are as follow:

1. Artifacts must now have a `task_path` and a `location` value where:
    * `task_path` is the path to the artifact in the temp folder when a task is run
    * `location`  is the name of the location where this artifact is persisted (can be an S3 bucket, a folder in a local filesystem or an HTTP endpoint to push the data to)

2. Artifact Location is a required value in workflows that manipulate artifacts. An artifact location is essentially a reusable config to facilitate individual artifact persistence. It also offers a layer of abstraction such that artifacts now only need to indicate where they are to be sourced/persisted relative to their Artifact Location and all the transfer mechanics (copy/paste for local file system, S3 transfer protocol for Buckets or HTTP calls for web API locations) is managed by the Artifact location object. Here is an example of the three Artifact Location types currently set up:

```yaml
artifact_locations:
- name: local-test
  type: local
  root: C:\\Users\Test\Projects\Project 1

- name: http-test
  type: http
  root: http://climate.onebuilding.org
  headers: 
    Authorization: some-long-JWT-token

- name: s3-test
  type: s3
  root: pollination
  endpoint: s3.eu-west-1.amazonaws.com
  bucket: all-of-my-data
  credentials_path: C:\\Users\Test\.queenbee\config.yaml
  # This config would result in the following S3 location
  # Get bucket "all-of-my-data" in region "eu-west-1" from AWS and 
  # only look for files with a path of  "pollination/*" in that bucket
```


AntoineDao:
Not sure why Travis is not running but all the tests passed! You can try it yourself if you don't believe me :stuck_out_tongue_closed_eyes:

AntoineDao:
In adb3d7e I also rewrote part of the workflow example to make it fit with the `radiance-folder` structure proposed in the repository with the same name.

It might not be perfect but the one thing to notice if you look at the workflow.yaml file without the diffs (to make it easier to read) is that all the parameter values used to help determine file paths have been converted to artifacts which can have values injected from the dag task to the function template if overriding is required.

queenbee/schema/function.py
```diff
@@ -134,24 +158,24 @@ def validate_variable(variables, func_name, input_names):
...
+                )
+                continue
             
             # now ensure input references are valid
             if ref != 'inputs' or typ != 'parameters':
                 continue
 
             assert name in input_names, \
-                'Illegal input parameter name in "%s": {{%s}}\n' \
+                'Illegal output parameter name in "%s": {{%s}}\n' \
```
mostaphaRoudsari: why output and should we change the next line to output too?

queenbee/schema/operator.py
```diff
@@ -109,11 +109,14 @@ class LocalRequirements(BaseModel):
     # NOTE: yaml conversion doesn't play well with Enum hence using a validator.
     @validator('platform')
     def check_platform(cls, v):
-        assert v.lower() in cls._valid_platforms, \
-            '{} is not a valid platform. Try one of these platforms:\n - {}'.format(
-                v, '\n - '.join(pl.title() for pl in cls._valid_platforms)
-            )
-        return v.lower()
+        v = list(map(lambda x: x.lower(), v))
```
mostaphaRoudsari: PEP8 👮 

```suggestion
        v = [x.lower() for x in v]
```

queenbee/schema/workflow.py
```diff
@@ -4,39 +4,63 @@
...
 from queenbee.schema.operator import Operator
 from queenbee.schema.function import Function
+from queenbee.schema.artifact_location import LocalLocation, HTTPLocation, S3Location
 
 
 class Workflow(BaseModel):
     """A DAG Workflow."""
-    type: Enum('Workflow', {'type': 'Workflow'}) = 'Workflow'
+
+    type: constr(regex='^workflow$')
```
mostaphaRoudsari: Was it always there and we didn't know? This is much cleaner.

queenbee/schema/workflow.py
```diff
@@ -4,39 +4,63 @@
...
+        description="A list of artifact locations which can be used by child flow objects"
+    )
 
+    # @validator('artifact_locations', each_item=True)
+    @root_validator
+    def check_references_exist(cls, values):
+        """Check that any artifact location referenced in templates or flows exists in artifact_locations"""
+        v = values.get('artifact_locations')
+        if v != None:
+            locations = list(map(lambda x: x.name, v))
```
mostaphaRoudsari:
```suggestion
            locations = [x.name for x in v]
```

queenbee/schema/workflow.py
```diff
@@ -4,39 +4,63 @@
...
+            artifacts = list_artifacts(values)
+            sources = list(set(map(lambda x: x.location, artifacts)))
```
mostaphaRoudsari:
```suggestion
            sources = list(set(x.location for x in artifacts))
```

queenbee/schema/workflow.py
```diff
@@ -77,6 +101,30 @@ def nodes_links(self):
...
+    if isinstance(obj, DAG):
+        for task in obj.tasks:
+            if 'arguments' in task and 'artifacts' in task.arguments:
+                artifacts = artifacts + task.arguments.artifacts  
+    
+    if isinstance(obj, Function):
+        if obj.inputs != None and obj.inputs.artifacts != None:
+            artifacts = artifacts + obj.inputs.artifacts
+        if obj.outputs != None and obj.outputs.artifacts != None:
+            artifacts = artifacts + obj.outputs.artifacts
```
mostaphaRoudsari: This will overwrite the input artifacts if any. Is this expected?

requirements.txt
```diff
@@ -1,2 +1,2 @@
-pydantic~=0.32.2
+pydantic~=1.0.0
```
mostaphaRoudsari: Happy you did this before it gets too late. 👏

tests/assets/artifact_locations.yaml
```diff
@@ -0,0 +1,17 @@
+artifact_locations:
+- name: local-test
+  type: local
+  root: C:\\Users\Test\Projects\Project 1
```
mostaphaRoudsari:
```suggestion
  root: C:\Users\Test\Projects\Project 1
```

tests/assets/artifact_locations.yaml
```diff
@@ -0,0 +1,17 @@
...
+  root: http://climate.onebuilding.org
+  headers: 
+    Authorization: some-long-JWT-token
+        
+- name: s3-test
+  type: s3
+  root: pollination
+  endpoint: s3.eu-west-1.amazonaws.com
+  bucket: all-of-my-data
+  credentials_path: C:\\Users\Test\.queenbee\config.yaml
```
mostaphaRoudsari:
```suggestion
  credentials_path: C:\Users\Test\.queenbee\config.yaml
```

tests/assets/function.yaml
```diff
@@ -11,5 +11,7 @@ command: gensky -c -B '{{ inputs.parameters.desired-irradiance }}' > '{{ inputs.
 outputs:
   artifacts:
     - name: sky
+      location: project-folder
       path: '{{ inputs.parameters.sky-file }}'
```
mostaphaRoudsari: So are we removing the quotes or not?

tests/assets/workflow_example/daylightfactor.yaml
```diff
@@ -64,104 +82,123 @@ templates:
...
+          location: project-folder
+          path: sky.file # Note that this is overwritten in the DAG by the output artifact path from the generate-sky task template
+          task_path: sky.file
+        - name: scene-files
+          location: project-folder
+          path: model
+          task_path: '.'
+    command: |
+      oconv \
+        ./static/aperture/*.mat ./static/aperture/*.rad \
```
mostaphaRoudsari: You can't pass * to a radiance command. What didn't you like about the initial set-up?

tests/assets/workflow_example/daylightfactor.yaml
```diff
@@ -64,104 +82,123 @@ templates:
...
 
 # this is not supported in Argo or is it?
-outputs:
-  artifacts:
-    - name: daylight-factor-values
-      path: "{{ tasks.merge-results.outputs.artifacts.result-file }}"
+# outputs:
+#   artifacts:
+#     - name: daylight-factor-values
+#       path: "{{ tasks.merge-results.outputs.artifacts.result-file }}"
```
mostaphaRoudsari: Why are we commenting this out?

tests/schema/artifact_location_test.py
```diff
@@ -0,0 +1,100 @@
...
+class ArtifactLocationReader(BaseModel):
+
+    artifact_locations: List[Union[LocalLocation, HTTPLocation, S3Location]]
+
+
+def test_create_local_location():
+    loc_dict = {
+        'name': 'local-test',
+        'type': 'local',
+        'root': 'C:\\\\Users\\Test\\Projects\\Project 1'
```
mostaphaRoudsari:
```suggestion
        'root': 'C:\\Users\\Test\\Projects\\Project 1'
```

tests/schema/artifact_location_test.py
```diff
@@ -0,0 +1,100 @@
...
+
+    assert obj == loc.to_dict()
+
+def test_load_artifact_locations():
+    locs = ArtifactLocationReader.from_file('./tests/assets/artifact_locations.yaml')
+
+    local_dict = {
+        'name': 'local-test',
+        'type': 'local',
+        'root': 'C:\\\\Users\\Test\\Projects\\Project 1'
```
mostaphaRoudsari:
```suggestion
        'root': 'C:\\Users\\Test\\Projects\\Project 1'
```

README.md
```diff
@@ -200,14 +231,16 @@ operator.
   operator: radiance-operator
   # commands and args will be used both locally and inside the container
   # TODO: This must change to be platform specific
-  command: gensky -c -B '{{ inputs.parameters.desired-irradiance }}' > '{{ inputs.parameters.sky-file }}'
+  command: gensky -c -B {{ inputs.parameters.desired-irradiance }} > file.sky
```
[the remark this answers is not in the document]
AntoineDao: From my understanding the way to use them is to encase an entire string which can contain templates in it. In the case above if you wanted to add quotes I would add them like this:

```yaml
  command: "gensky -c -B {{ inputs.parameters.desired-irradiance }} > file.sky"
```

Otherwise it will mess with yaml printing.

queenbee/schema/function.py
AntoineDao: Oh that's my bad. I think the check run by this function applies to both inputs and outputs. Think we can leave it for now and raise an issue to generally improve messaging/debugging.

queenbee/schema/operator.py
AntoineDao: PEP8 doesn't allow map functions?!

queenbee/schema/workflow.py
AntoineDao: Yep... :smile:

tests/assets/function.yaml
AntoineDao: Missed that one. I would double quote it :smile:

queenbee/schema/workflow.py
AntoineDao: Yep, artifacts is just an empty list created at the start of this function. It shouldn't change anything in the schema (unless I'm misreading my own code here).

tests/assets/workflow_example/daylightfactor.yaml
AntoineDao: Didn't really have any qualms with the initial setup, although it wasn't super clear what the artifacts were (were you passing a folder or a file?).

With regards to this command, I have no idea what radiance can/cannot accept, I just blindly copy/pasted this: https://github.com/ladybug-tools/radiance-folder-structure/tree/master/project_folder/model#ray-tracing-with-rtrace--rcontrib--rpict

tests/assets/workflow_example/daylightfactor.yaml
AntoineDao: That's my bad! I left it out because it was confusing me during the testing process. Forgot to uncomment.

tests/assets/workflow_example/daylightfactor.yaml
AntoineDao: Resolved it here: d3b8a55 

Feel free to re-open if not happy with it.

queenbee/schema/operator.py
mostaphaRoudsari: lambda is the issue here but generally speaking I think list comprehension is much cleaner in this case.

queenbee/schema/workflow.py
mostaphaRoudsari: but if inputs and outputs both have artifacts don't you want to collect them all?

```python
        artifacts = []
        if obj.inputs != None and obj.inputs.artifacts != None:
            artifacts.extend(artifacts + obj.inputs.artifacts)
        if obj.outputs != None and obj.outputs.artifacts != None:
            artifacts.extend(artifacts + obj.outputs.artifacts)
```

PR #13, opened 2019-11-09
Cli

AntoineDao:
This PR implements a basic CLI using click and adds graph visualization capabilities through graphviz.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -8,3 +8,4 @@ sphinx-bootstrap-theme==0.6.5
 sphinxcontrib-fulltoc==1.2.0
 sphinxcontrib-websupport==1.1.0
 twine==1.13.0
+click>=5.1
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -8,4 +8,4 @@ sphinx-bootstrap-theme==0.6.5
 sphinxcontrib-fulltoc==1.2.0
 sphinxcontrib-websupport==1.1.0
 twine==1.13.0
-click>=5.1
+click>=7.0
--- a/queenbee/cli.py
+++ b/queenbee/cli.py
@@ -61,6 +61,8 @@ except ImportError:
         'click module is not installed. Try `pip install queenbee[cli]` command.'
     )
 
+import webbrowser
+import urllib.parse
 from queenbee.schema.workflow import Workflow
 
 # TODO: Comment out and use for logging once we are adding commands to this file.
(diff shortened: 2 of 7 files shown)


dev-requirements.txt
```diff
@@ -8,3 +8,4 @@ sphinx-bootstrap-theme==0.6.5
 sphinxcontrib-fulltoc==1.2.0
 sphinxcontrib-websupport==1.1.0
 twine==1.13.0
+click>=5.1
```
mostaphaRoudsari: Should we use a more recent version? The latest release is version 7.0.

queenbee/cli.py
```diff
@@ -0,0 +1,136 @@
...
+    ctx.obj = Context()
+    
+
+@main.command('viz')
+def viz():
+    """Check if queenbee is flying!"""
+    click.echo("""
+    
+                                       .' '.            __
+  viiiiiiiiiiiiizzzzzzzzz!  . .        .   .           (__\_
```
mostaphaRoudsari: Super! 🐝

queenbee/cli.py
```diff
@@ -0,0 +1,136 @@
...
+                               .         .         . -{{_(|8)
+                                ' .  . ' ' .  . '     (__/
+
+    """)
+
+@main.command('validate')
+@click.option('-f', '--file')
+@click.pass_context
+def validate(ctx, file):
```
mostaphaRoudsari: You might want to add an option to avoid using GaphvizOnline on validation. This might get annoying after some time.

Or maybe this can be two different commands. The visualize command can validate it first under the hood but not the other way around.

queenbee/schema/workflow.py
```diff
@@ -88,6 +89,19 @@ def visualize(self):
         # call index.html
         os.system(os.path.join(directory, 'chart/index.html'))
 
+    def to_diagraph(self, filename=None):
+        """Return a graphviz instance of a diagraph from workflow"""
+        if filename is None:
+            filename = self.id
+        f = Digraph(self.name, filename='{}.gv'.format(filename))
```
mostaphaRoudsari: Now that we are using this library should we remove the other visualization option?

queenbee/cli.py
AntoineDao: Yep you're right. I fixed it by adding an "--display" flag option

queenbee/schema/workflow.py
AntoineDao: I think so.

PR #15, opened 2019-11-11
Workflow hydration

AntoineDao:
I added this is as I was encountering issues with Argo whereby I wanted to set the path of an artifact to be `argo/pollination/job/{{ workflow.id}}`.

I could have written these methods in the other repo but this felt like the kind of thing we should add to queenbee.

I also added some minor fix to pydantic enums (deleted them :stuck_out_tongue: ) and added one-line doc to the `validate` command.

### DIFF
--- a/queenbee/schema/workflow.py
+++ b/queenbee/schema/workflow.py
@@ -6,6 +6,7 @@ end steps for the workflow.
 from uuid import UUID, uuid4
 import json
 import os
+import re
 from graphviz import Digraph
 from pydantic import Field, validator, constr, root_validator
 from typing import List, Union
@@ -15,6 +16,7 @@ from queenbee.schema.arguments import Arguments
 from queenbee.schema.operator import Operator
 from queenbee.schema.function import Function
 from queenbee.schema.artifact_location import LocalLocation, HTTPLocation, S3Location
+from queenbee.schema.parser import parse_double_quote_workflow_vars, replace_double_quote_vars
 
 
 class Workflow(BaseModel):
@@ -48,7 +50,6 @@ class Workflow(BaseModel):
         description="A list of artifact locations which can be used by child flow objects"
     )
 
-    # @validator('artifact_locations', each_item=True)
     @root_validator
     def check_references_exist(cls, values):
         """Check that any artifact location referenced in templates or flows exists in artifact_locations"""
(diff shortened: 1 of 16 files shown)


tests/assets/workflow_example/daylightfactor.yaml
```diff
@@ -167,7 +167,7 @@ flow:
       arguments:
         parameters:
           - name: scene-file-paths
-            value: "{{ workflow.inputs.parameters.scene-file-paths }}"
+            value: "{{ workflow.inputs.parameters.scene-file-paths.value }}"
```
mostaphaRoudsari: Interesting! What are the other properties here that one can use?

queenbee/schema/workflow.py
```diff
@@ -84,6 +87,33 @@ def to_diagraph(self, filename=None):
...
+                else:
+                    obj = obj[key]
+
+            values[ref] = obj
+
+        return values
+
+
+    def hydrate_workflow_templates(self):
+        """returns a copy of the workflow but the {{workflow.x.y.z}} templates are replaced with the actual values referenced}}"""
```
mostaphaRoudsari: This line is too long! PEP8 🚨

queenbee/schema/workflow.py
```diff
@@ -97,6 +127,50 @@ def nodes_links(self):
         return {'nodes': nodes, 'links': links}
 
 
+def hydrate_templates(workflow, wf_value=None):
+    """cyle through an arbitary workflow value (dictionary, list, string etc...) and hydrate any workflow template value with it's actual value"""
```
mostaphaRoudsari: This line is too long! PEP8 🚨

queenbee/schema/workflow.py
```diff
@@ -97,6 +127,50 @@ def nodes_links(self):
         return {'nodes': nodes, 'links': links}
 
 
+def hydrate_templates(workflow, wf_value=None):
```
mostaphaRoudsari: This is nice! So you are basically making a static version of the workflow with values provided by user. I think it will be helpful to add more description on why or when people should use this.

tests/assets/workflow_example/daylightfactor.yaml
AntoineDao: At this stage the function fetching values recurses through the entire object looking for `{{ workflow.x.y...z }}` so pretty much anything in the workflow object.

queenbee/schema/workflow.py
AntoineDao: Will do. My thinking was that it would mostly be used by the plugins when writing/running workflows.

AntoineDao:
I'll install a PEP8 plugin into vscode just now... :police_car:

PR #18, opened 2019-11-13
Revert pydantic version

AntoineDao:
This was not an enjoyable process... Glad we've got tests to provide some level of confidence in the fact I haven't broken everything. 

I'm going to go ahead and merge this just now @mostaphaRoudsari so that I can keep working on workflow-service. Feel free to open an issue if you're wanting to revert this revert :upside_down_face:


mostaphaRoudsari:
Oh man this is super sad! I saw that there is a PR for version 1.0 but it will take some time before it will be available. Hopefully it will be as simple to go back to version 1.0.

AntoineDao:
Yeah I know... Should be relatively simple to `un-revert` (famous last words :upside_down_face: )

PR #23, opened 2020-01-07
fix(function): support integer inputs

mostaphaRoudsari:
- Fixed the validator to allow integer inputs for arguments.
- Made several PEP8 fixes and corrected a number of misspellings.


AntoineDao:
I'll be rude and merge it for you just now so we can bump queenbee versions in other repos.

PR #24, opened 2020-01-07
docs(workflow): add workflow documentation

mostaphaRoudsari:
Since right now the documentation is mainly for the workflow itself and not the python client-side libraries I updated the documentation to generate an OpenAPI compatible documentation for Queenbee schema which will be rendered by redoc. The changes are very similar to the one in honeybee-model-schema repository.


queenbee/_openapi.py
```diff
@@ -0,0 +1,98 @@
...
+from typing import Dict
+
+from .schema.workflow import Workflow
+
+_base_open_api = {
+    "openapi": "3.0.2",
+    "servers": [],
+    "info": {
+        "description": "This is the documentation for Queenbee Workflow schema.",
+        "version": "1.7.0",
```
AntoineDao: Any way we can set this based on the library version rather than have it hard coded? I'm thinking we might be able to do so by running the "gh-pages" job as a github action triggered by a release event.

mostaphaRoudsari: I was lazy to do it right. It is now looking up for the version.

PR #30, opened 2020-01-18
Improved validators and documentation

### LINKED ISSUE #11: Add check for operators existing
Currently any operator name can be typed in for each task/template. This should be constrained to the operators declared in the `operators` section of a workflow.

### LINKED ISSUE #12: Template parameters with double quotes
Seems we've been a bit all over the place with our use of single quotes, double quotes or no quotes for `{{template}}` values.

### LINKED ISSUE #19: Check dependencies exist in DAG
We should check that a dag task's dependencies actually existing within the list of task templates defined.

mostaphaRoudsari:
- Added new validators
- Fixed the bug for finding workflow artifacts.
- Added new methods to Argument, Parameter and Artifact classes to make it easy to access and validate the input values
- Moved input validation to Arguments level so we can validate them before adding them to the workflow.
- Added new tests for Parameter.
- Added a new module for queenbee variables.

This PR partially breaks workflow hydrate which I will fix in the next PR.

resolves #12, resolves #19, resolves #11

### DIFF
--- a/queenbee/schema/arguments.py
+++ b/queenbee/schema/arguments.py
@@ -9,9 +9,10 @@ Queenbee accepts two types of arguments:
     2. Artifact: An ``artifact`` is a file or folder that can be identified by a url or
         a path.
 """
-from queenbee.schema.qutil import BaseModel
+from queenbee.schema.qutil import BaseModel, find_dup_items
 from queenbee.schema.artifact_location import VerbEnum
-from pydantic import Field
+import queenbee.schema.variable as qbvar
+from pydantic import Field, root_validator
 from typing import List, Any, Optional, Dict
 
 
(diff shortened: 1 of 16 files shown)


queenbee/schema/arguments.py
```diff
@@ -40,22 +41,69 @@ class Parameter(BaseModel):
...
         ...,
-        description='name of the artifact. must be unique within a task\'s '
+        description='Name of the artifact. Must be unique within a task\'s '
         'inputs / outputs.'
     )
 
     location: str = Field(
-        None,
-        description="Name of the Artifact Location to source this artifact from."
+        None,  # is it possible to create an artifact with no artifact location?
```
AntoineDao: I'm not sure... My instinct tells me no...

queenbee/schema/workflow.py
```diff
@@ -200,31 +294,6 @@ def hydrate_templates(workflow, wf_value=None):
     return wf_value
 
 
-def list_artifacts(obj):
```
AntoineDao: Why did we remove this?

tests/assets/function.yaml
```diff
@@ -7,11 +7,11 @@ inputs:
     - name: sky-file
       description: full path to output sky file
 operator: radiance-operator
-command: gensky -c -B '{{ inputs.parameters.desired-irradiance }}' > '{{ inputs.parameters.sky-file }}'
+command: gensky -c -B "{{ inputs.parameters.desired-irradiance }}" > "{{ inputs.parameters.sky-file }}"
```
AntoineDao: My understanding is that if `desired-irradiance=5` and `sky-file=file.sky` this will print as `gensky -c "5" > "file.sky"`. Is this the behavior we expect?

mostaphaRoudsari: My undrstanding is that it will be `gensky -c 5 > file.sky`. :)

Double quotes are required to make the variable being recognized from a dictionary. See here: https://github.com/argoproj/argo/tree/711bb11483a0ccb46600795c636c98d9c3a7f16c/examples#parameters

image

queenbee/schema/workflow.py
mostaphaRoudsari: I implemented it for each object. For instance `workflow.artifacts` returns the artifacts in that workflow and so on.

queenbee/schema/arguments.py
mostaphaRoudsari: Thanks. I will make it a required input then.

queenbee/schema/workflow.py
AntoineDao: Aaaah I see it now. Neat!

tests/assets/function.yaml
AntoineDao: I'm not sure that's the case if it's a template value within a string. ie: 'I really like {{something.i.like}}' doesn't need `"` because it's already within a string but `["{{something}}"] does because the whole template value will be a string.

Doesn't matter too much for now. Let's see how it pans out and worry about it when it breaks.

PR #34, opened 2020-01-23
Add more validations to workflow

mostaphaRoudsari:
I'm sending this PR mainly to show the process to @MingoPeng but I also wanted the last changes to be part of the tests that @AntoineDao is doing on the simulation service.

### DIFF
--- a/queenbee/schema/arguments.py
+++ b/queenbee/schema/arguments.py
@@ -246,6 +246,28 @@ class Arguments(BaseModel):
         """Get an artifact value by name."""
         return self.get_artifact(name).current_value
 
+    def set_parameter_value(self, name, value):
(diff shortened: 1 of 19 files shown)


queenbee/schema/arguments.py
```diff
@@ -164,7 +166,7 @@ def current_value(self):
         return self.path if self.path is not None else self.source_path
 
     @property
-    def ref_vars(self) -> Dict[str, List[str]]:
+    def referenced_values(self) -> Dict[str, List[str]]:
```
MingboPeng: this looks good.

mostaphaRoudsari: Thanks!

PR #37, opened 2020-02-18
Add new fields for info and user_data

mostaphaRoudsari:
Info closely follows the openapi info object. It allows users to add description, version, license and contact information to a workflow. A couple of notes:

- I kept the name of the workflow as is not to make a breaking change. It might be a better implementation to move it under info.
- All the field are currently optional but we might want to make versioning required.
- Should we consider a license by default? Something like MIT or BSD?

I added user_data field separate from the other arguments input since I _feel_ it serves a different purpose and should be on its own. The current implementation accepts a dictionary and is optional.

### DIFF
--- a/queenbee/schema/arguments.py
+++ b/queenbee/schema/arguments.py
@@ -191,6 +191,9 @@ class Arguments(BaseModel):
     Queenbee accepts two types of arguments: parameters and artifacts. A ``parameter``
     is a variable that can be passed to a task or a workflow. An ``artifact`` is a file
     or folder that can be identified by a url or a path.
+
+    There is a third field for ``user_data`` which can be used to assign arbitrary values
+    to a workflow.
     """
 
     parameters: List[Parameter] = Field(
@@ -204,6 +207,12 @@ class Arguments(BaseModel):
         None,
         description='Artifacts is the list of file and folder arguments to pass to the '
         'task or workflow.'
+    )
+
+    user_data: Dict = Field(
+        None,
+        description='Optional user data as a dictionary. User data is for user reference'
+        ' only and will not be used in the execution of the workflow.'
     )
 
     @root_validator
--- a/queenbee/schema/arguments.py
+++ b/queenbee/schema/arguments.py
@@ -191,9 +191,6 @@ class Arguments(BaseModel):
     Queenbee accepts two types of arguments: parameters and artifacts. A ``parameter``
     is a variable that can be passed to a task or a workflow. An ``artifact`` is a file
     or folder that can be identified by a url or a path.
-
-    There is a third field for ``user_data`` which can be used to assign arbitrary values
-    to a workflow.
     """
 
     parameters: List[Parameter] = Field(
@@ -207,12 +204,6 @@ class Arguments(BaseModel):
         None,
         description='Artifacts is the list of file and folder arguments to pass to the '
         'task or workflow.'
-    )
-
-    user_data: Dict = Field(
-        None,
-        description='Optional user data as a dictionary. User data is for user reference'
-        ' only and will not be used in the execution of the workflow.'
     )
 
     @root_validator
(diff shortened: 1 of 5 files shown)


mostaphaRoudsari:
@AntoineDao, what do you think about adding a column to `Backend` board for review? Didn't know where to put this PR.

queenbee/schema/arguments.py
```diff
@@ -206,6 +209,12 @@ class Arguments(BaseModel):
         'task or workflow.'
     )
 
+    user_data: Dict = Field(
```
AntoineDao: Can we call this `labels` or `tags` rather than user data? If the intent is for them to be used as reference material I think those works ,make more sense.

queenbee/schema/info.py
```diff
@@ -0,0 +1,64 @@
...
+        description='The license name used for the workflow.'
+    )
+
+    url: str = Field(
+        None,
+        description='A URL to the license used for the workflow.'
+    )
+
+
+class Info(BaseModel):
```
AntoineDao: Can we just call this `Metadata`?

queenbee/schema/workflow.py
```diff
@@ -30,8 +31,14 @@ class Workflow(BaseModel):
 
     id: str = str(uuid4())
 
+    info: Info = Field(
```
AntoineDao: re: calling this item `metadata`

queenbee/schema/arguments.py
AntoineDao: Also, I understand what we're trying to achieve with this in the sense that an individual simulation should be able to label itself with additional metadata. However it really rubs me the wrong way that we would put this in the `arguments` schema. This is because arguments are used everywhere in both input, outputs, individual task template etc... Could we put this at the workflow level and make it another variable that a `simulation trigger` payload would impact?

queenbee/schema/workflow.py
mostaphaRoudsari: We can! I didn't want to create a conflict if the key is used by K8. Otherwise I think `metadata` is the better choice.

queenbee/schema/arguments.py
mostaphaRoudsari: The reason I put it here is that one can update the values in an argument using an input file. Maybe I should create a separate type only for workflow inputs which can have `user_data` field.

I thought we should keep the tag as a list for workflow. For instance it will be daylighting, energy, etc. `user_data` has a different use case than tags. I'm down with using labels which will match k8 terminology I guess.

queenbee/schema/workflow.py
AntoineDao: None of this information should make it's way into k8s right?

queenbee/schema/arguments.py
mostaphaRoudsari: I added a new object type for WorkflowArgument which will only be used as a workflow input. It has an extra field for user_data which is a dictionary.

I will add tags and remove the default value for id in a different PR.

mostaphaRoudsari:
ping @AntoineDao!

mostaphaRoudsari:
Hi @AntoineDao, Thanks. This is actually how it is designed in Rhino and I intentionally stayed away from it as it made a lot of headache for using it in Honeybee. For instance with this strategy you can't have nested dictionaries as user_data. But that was my personal experience. Curious to know why do you think we should limit the types?

AntoineDao:
I'm not fussed either way. Was just saying that if we don't want nested JSON we should enforce a type. If we're expecting nested JSON then let's keep it as is.

theo-armour:
With Three.js .userData, it seems that "anything goes" is the policy and I've not heard of any significant issues arising from this policy. So, as an ignorant bystander, I suggest a similar path here.

And I am happy with an "anything goes" MIT license as well;

mostaphaRoudsari:
Thank you @AntoineDao and @theo-armour for your input. I will leave it as is for now and we will see if there will be any issues in the future.

queenbee/schema/metadata.py
```diff
@@ -48,9 +49,9 @@ class MetaData(BaseModel):
         description='A short description of the workflow.'
     )
 
-    contact: Contact = Field(
+    author: Union[Author, List[Author]] = Field(
```
AntoineDao: Oh hadn't caught this when reviewing your latest changes. Can we just have it as one type: `List[Author]`? I'm keen to reduce unions to where they are absolutely necessary to reduce overall schema complexity.

mostaphaRoudsari: Sounds good. I will push a fix!

PR #44, opened 2020-03-30
Test refractor

AntoineDao:
The changes proposed in the PR essentially refractor a lot of the re-usable tests into test classes that can parametrize the fixtures based on the assets folder.

A lot of asset files have been moved around using the following logic:
* **Correct Schema Objects**: `tests/assets/<ClassName>/valid/`
* **Incorrect Schema Objects**: `tests/assets/<ClassName>/invalid/` where you can optionally add a file with the same name as your test asset but ending in `.error` containing the error message you expect to appear

I have refractored/removed tests that have been duplicated as a result of adding the test classes. I have left all the other class specific tests such as workflow hydration tests.

PS: I have made this PR separately to other refractor work in order to keep the reviewer's mind (:wink: @mostaphaRoudsari :wink:) fresh when looking over this PR.

PR #45, opened 2020-03-30
Workflow variables

AntoineDao:
This PR adds a root_validator for the workflwo schema that checks any referenced values are valid. It also resolves these values during workflow hydration.

This PR should be merged after https://github.com/ladybug-tools/queenbee/pull/44 as I have merged the test refractors within it. (not best practice but makes it easier to test code while I wait for @mostaphaRoudsari to review both PRs).


queenbee/schema/artifact_location.py
```diff
@@ -9,11 +9,28 @@
...
+        return ref_values
+
+    for value in values:
+        if value is None:
+            continue
+        ref_var = qbvar.get_ref_variable(value)
+        if ref_var:
+            ref_values[value] = ref_var
+
+    return ref_values
```
mostaphaRoudsari: I think we should either make this function  hidden and use it only for internal use in this module or move it to a helper module.

mostaphaRoudsari: Maybe even add it as a static method to the BaseModel and reference it from these classes?

queenbee/schema/workflow.py
```diff
@@ -422,6 +472,56 @@ def hydrate_templates(workflow, wf_value=None):
     return wf_value
 
 
+def assert_workflow_values(values, references):
```
mostaphaRoudsari: Similar to the other case, since this is for internal use only we should either make it internal or make it  a `staticmethod` and part of the workflow class itself.

queenbee/schema/workflow.py
```diff
@@ -422,6 +472,56 @@ def hydrate_templates(workflow, wf_value=None):
...
+        if not ref.startswith('workflow.'):
+            continue
+
+        try:
+            _, attr, prop, name = ref.split('.')
+        except ValueError:
+            *_, name = ref.split('.')
+
+            assert name in ['id', 'name'], ValueError(
+                f'Invalid workflow variable: {ref}',
```
mostaphaRoudsari: We should add a longer explanation on why it is invalid.

queenbee/schema/arguments.py
```diff
@@ -293,7 +293,29 @@ class WorkflowArguments(Arguments):
     field for user_data.
     """
     user_data: Dict = Field(
-        None,
+        {},
```
mostaphaRoudsari: `None` is the better option here. Using mutable objects as default argument values can be tricky:

https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

queenbee/schema/arguments.py
```diff
@@ -293,7 +293,29 @@ class WorkflowArguments(Arguments):
...
+    )
+
+
+class WorkflowInputs(BaseModel):
+    """Inputs to inject into a workflow before hydration
+    
+    """
+
+    artifacts: Dict[str, dict] = Field(
+        {},
```
mostaphaRoudsari:
```suggestion
        None,
```

queenbee/schema/arguments.py
```diff
@@ -293,7 +293,29 @@ class WorkflowArguments(Arguments):
...
+        description='A dictionary containing the workflow input argument name as a key and a dictonary of key value pairs to update'
+    )
+
+    parameters: Dict[str, dict] = Field(
+        {},
```
mostaphaRoudsari:
```suggestion
        None,
```

queenbee/schema/arguments.py
```diff
@@ -293,7 +293,29 @@ class WorkflowArguments(Arguments):
...
+        description='A dictionary containing the workflow input argument name as a key and a dictonary of key value pairs to update'
+    )
+
+    user_data: Dict = Field(
+        {},
```
mostaphaRoudsari:
```suggestion
        None,
```

queenbee/schema/qutil.py
```diff
@@ -24,7 +24,7 @@ class BaseModel(PydanticBaseModel):
 
     def yaml(self, exclude_unset=False):
         return yaml.dump(
-            self.dict(exclude_unset=exclude_unset),
+            json.loads(self.json(exclude_unset=False)),
```
mostaphaRoudsari: 1. Why are we set exclude_unset to False? I found it helpful to set it to True when I want to generate clean YAML files.
2. If you think it should set to False all the time then we have to remove the argument in line 25.

queenbee/schema/arguments.py
```diff
@@ -293,7 +293,29 @@ class WorkflowArguments(Arguments):
...
     """
     user_data: Dict = Field(
-        None,
+        {},
+        description='Optional user data as a dictionary. User data is for user reference'
+        ' only and will not be used in the execution of the workflow.'
+    )
+
+
+class WorkflowInputs(BaseModel):
```
mostaphaRoudsari: What's the difference between `WorkflowInputs` and `WorkflowArguments`? Why do we need this one?

queenbee/schema/qutil.py
AntoineDao: This is a typo on my part. Will fix it.

queenbee/schema/arguments.py
AntoineDao: Workflow Inputs are the payload sent to the user to modify the Workflow arguments. In your initial implementation your proposed the following schema which does not match the Worfklow arguments so I formalised it:

```
        The dictionary should have two keys for ``parameters`` and ``artifacts``.
        The values for each key is a dictionary where the key is the name of the
        target variable and value is ``key: value`` that should be updated.

        For instance this input updates the value for workers to 6 and grid-name to
        classroom.

        .. code-block:: python

            {
                'parameters': {
                    'worker': {'value': 6},
                    'grid-name': {'value': 'classroom'}
                }
            }

```

queenbee/schema/arguments.py
AntoineDao: Hmm I see your point. The reasoning behind this was to make the default workflow as complete as possible to avoid having to run endless `if "user_data" in workflow.arguments: etc...` It feels like an ok sacrifice to make the code less verbose and offload the correct `parsing` of objects to Pydantic.

PR #49, opened 2020-05-06
Tuesday Update

AntoineDao:
The changes included in today's PR involve:
* Fixing some validation for recipes and baked recipes
* Enabling creation of workflows from a baked recipe (ie: Workflow = Baked Recipe + Arguments)
* Update the testing assets to reflect working setup


tests/assets/recipes/parametric-daylight-factor/example-arguments.yaml
```diff
@@ -0,0 +1,13 @@
+artifacts:
+  - name: models-folder
+    source:
+      type: project-folder
+      path: path/to/models
+parameters:
+  - name: model-names
+    value: |
```
mostaphaRoudsari, line 8: fancy!

tests/assets/recipes/parametric-daylight-factor/flow/main.yaml
```diff
@@ -0,0 +1,47 @@
...
+    from:
+      type: inputs
+      variable: model-names
+  arguments:
+    artifacts:
+    - name: model
+      from:
+        type: inputs
+        variable: models-folder
+      subpath: '{{item}}/model'
```
mostaphaRoudsari, line 24: nice trick here! Is it supported for all artifacts? Does it mean that this artifact will always be copied under this subfolder?

tests/assets/repository/index.yaml
```diff
@@ -0,0 +1,60 @@
+{
+  "generated": "2020-05-06T20:30:58.581950",
+  "operator": {
+    "honeybee-radiance": [
+      {
+        "name": "honeybee-radiance",
+        "version": "1.2.3",
+        "appVersion": "2.3.2",
```
mostaphaRoudsari, line 8: `app-version` ?

tests/assets/repository/index.yaml
```diff
@@ -0,0 +1,60 @@
...
+        ],
+        "home": null,
+        "sources": null,
+        "icon": "https://www.ladybug.tools/assets/img/logo.png",
+        "deprecated": null,
+        "description": "An awesome workflow to run daylight factor!",
+        "license": null,
+        "url": "recipes/daylight-factor-0.0.1.tgz",
+        "created": "2020-05-06T20:30:58.581874",
+        "digest": "dfe964ce20bca8a1685b52cfce2b810d10cc31fb9ac15daa1128f3d712228d48"
```
mostaphaRoudsari, line 56: interesting! how does one provide this?

tests/assets/repository/index.yaml
AntoineDao: Picked it up from helm. The Operator version corresponds to the version of the YAML file + version of container or library. You might add a new function but not change the version of the underlying package/container. In which case you indicate what the "appVersion".

tests/assets/repository/index.yaml
AntoineDao: Automatically generated. The index is generated from commands by either reading a folder or packaging and adding a new operator/recipe to a repository folder.

AntoineDao: The idea here is to make sure that if someone overwrites a version somehow, any dependencies are aware of it and can decide to either accept the changes or have their recipe broken. It's WIP but a powerful concept.

tests/assets/recipes/parametric-daylight-factor/flow/main.yaml
AntoineDao: This is the inputs artifacts. This was used as syntactic sugar to make it possible to set paths dynamically through a parameter. It is mostly useful for loops as demoed above.

tests/assets/recipes/parametric-daylight-factor/example-arguments.yaml
AntoineDao: Meh... Does the trick. I'd be happier with actual JSON but we'll get there later :)

tests/assets/repository/index.yaml
mostaphaRoudsari: Thanks. My comment was mainly about the semantics. I think we should be consistent on keywords and not use camelCase here! 😁

AntoineDao: Oh! Hadn't thought about it in that way. I'll create an issue to tackle this in a coordinated manner as there are probably a couple more lying around! https://github.com/ladybug-tools/queenbee/issues/50

PR #53, opened 2020-05-13
Documentation!

AntoineDao:
This PR adds docstrings and rewrites existing documentation concerning how to create recipes/operators and use the queenbee CLI.


queenbee/base/parser.py
```diff
@@ -5,20 +5,24 @@
 import re
 
 
-def _check_list(lst, folder):
+def _check_list(lst: list, folder: str):
+    # TODO: Ask Mostapha for docs
```
mostaphaRoudsari, line 9: Mostapha has no idea right now! He can add it later when starts using the library. 😁

queenbee/recipe/recipe.py
```diff
@@ -65,6 +72,30 @@ class Recipe(BaseModel):
...
+        Note:
+            Here is an example of a Recipe folder:
+            ::
+
+                .
+                ├── .dependencies
+                │   ├── operators
+                │   │   └── <sha-256>.yaml
+                │   └── recipes
+                │       └── <sha-256>.yaml
```
mostaphaRoudsari, line 86: I assume this will be generated by a command? Do we have some documentation for this. I like the idea of how you are keep tracking on the unique items. 👍

AntoineDao: All part of the joys of building a package manager :upside_down_face:

AntoineDao:
@mostaphaRoudsari I will leave this for today. I will add some schema examples tomorrow and then merge if that's ok? Or we can merge this now and I can add workflow and operator examples in another PR if that's more readable for you.

mostaphaRoudsari:
I prefer to merge it now.

I understand that since this is still a work in progress we want to keep the documentation short but once you add the examples we need to include a short writing in README.md. Otherwise it will be really hard for people to understand what is this all about. 😀

Issue #47, opened 2020-05-04
Package and Distribute Operators and Recipes like Helm


AntoineDao:
Helm has a robust packaging and distribution strategy for Charts that relies on generating yam indices.

Here is a [good link to their docs](https://helm.sh/docs/topics/chart_repository/) to give an idea of what I am talking about.

AntoineDao:
Resolved in #61

Issue #42, opened 2020-03-29
Hydrate workflow template inside a workflow


AntoineDao:
Workflows make use of string referenced values (eg: `{{workflow.inputs.parameter.foo}}`) to enable re-usability. A workflow is `hydrated` before it is scheduled to run. During this process all the `workflow` level referenced values are resolved.

This works fine in most cases however I don't think the current code handles hydrating a workflow template within a workflow very well.

Here are the workflow values that won't resolve well:

* Artifact Location: If a workflow is within another flow then it's `input-folder` location can be either the parent workflow's `input-folder` or it should be set explicitely through the `flow` task input artifacts
* Workflow level referenced values: It's unclear if they should be resolved with the child or parent workflow values. We may want to resolve both at the same time somehow?

AntoineDao:
Resolved in #61

Issue #39, opened 2020-02-18
Recognize the difference between a workflow and a simulation


mostaphaRoudsari:
This is related to #32 and how hydrating workflows is currently done. Some notes about the terminologies:

## Workflow
A workflow is reusable. It means you can use the workflow over and over by just changing the arguments and the logic stays the same. Workflow is useful to define the process.

## Simulation (for the lack of a better term)
A simulation is a hydrated/frozen version of a workflow. It has assigned values for each input argument which is passed to the workflow in the time that the simulation is created. Also there is no referenced values in a simulation. All the referenced values has been replaced by input values. Simulation is useful to execute the process.

I think we should recognize and respect this difference or otherwise we have two very similar object which will be hard to recognize from each other.

With this logic one should not be able to hydrate a workflow. You can get a hydrated version of the workflow at any point but that should be called something else like `simulation`.

@AntoineDao, I mainly opened this issue to get your feedback on this:

1. Do you agree with this distinction?
2. If yes, what would you call the second one? Should it be a new schema of its own that for example doesn't have assignable inputs?

mostaphaRoudsari:
We can also rename what we call `workflow` now to `recipe` and call a hydrated-recipe a `workflow`.

A recipe is a set of instructions with inputs and referenced values to inputs while a workflow is ready-to-be-executed and has no referenced values.

theo-armour:
workflow versus simulation versus ???

With these types of situations, I like to follow the lead of whatever tools
I am working with.

Examples I have dealt with myself

Three.js theta and phi
Mapping: latitude and longitude
Solar calcs: azimuth and bearing

So when dealing with Radiance, copy Radiance terminology. Ditto Dynamo,
Grasshopper.

Do not imagine that peeps will instantly love concepts you just made up one
night. Copy others.

In other words: "Don't make me think", "Don't make me change."

AntoineDao:
Resolved in #61

PR #61, opened 2020-05-19
Queenbee Refactor

AntoineDao:
# Overview
This PR contains the changes made during April/May 2020 to refactor Queenbee. The following major changes were made:

# Warning
I think it's best we **do not** mark this PR as `Breaking Change`. I know it does introduce breaking changes, however we are already at `v1.0.0`. I would rather say that the `1.0` version is unstable and prone to change until we deem is ready instead of regularly bumping major versions while we get Queenbee up and going in earnest.

# Details
## Separate `Workflow` into `Operators`, `Recipes` and `Workflows`
Two main issues led us to split the old `Workflow` object into three separate objects:

* Confusion between Workflow template and Workflow object executed by Argo or Luigi #39
* Re-usability of Workflows

For this reason we introduce the three following Queenbee Objects:

* **Operator:** A list of `functions` (templated commands) for 1 cli tool (eg: `radiance-operator` would template command for radiance)
* **Recipe:** A reusable DAG definition of Operator Functions or other Recipes (#42) (ie: a recipe can use another recipe inside of itself)
* **Workflow:** A Recipe + Input Arguments run by an executor (ie: luigi or argo). The schema will not be used by Queenbee explicitely but serves as a common definitions or interface for the different execution engines to report progress of an executed Recipe.

## Introduce Dependency Management
Despite [all warnings](https://medium.com/@sdboyer/so-you-want-to-write-a-package-manager-4ae9c17d9527) we went ahead and built a package manager... This was a requirement for `Recipes` to be shareable and modular (ie: a recipe could refer to a certain operator without having to copy all the operator data).

To do so we drew heavy inspiration from the [Helm project](https://helm.sh/docs/topics/chart_repository/) (#47 ). This means that Queenbee packaging and sharing remains open sourced and freely available for those who wish to create their own registries of Recipes and Operators.

## Artifact Path Context (remove `artifact_location`)

Artifact Locations have been removed to make recipes more reusable. Hard coding the location to persist a file inside a DAG lead to re-usability issues when nesting Recipes inside each other. It was therefore preferable to establish the concept of `artifact path context`:

Queenbee Workflow - Cloud Path Context

Queenbee Workflow - Local Run Path Context

### Function Path Context
When an artifact path is declared within an `Operator Function`, this path is only relevant or valid in the context of this function. In practice this means that a Function doesn't care where an artifact is persisted when it is executed because it knows that the executor (argo or luigi) will move copy the artifact to the path it expects it to be in the command.

```yaml
name: ray-tracing
description: Run ray tracing using some input data!
inputs:
  parameters:
  - name: radiance-parameters
    description: a string of radiance parameters
    default: -b 5
  artifacts:
  - name: grid
    path: grid.pts
  - name: scene-file
    path: scene.oct
command: rtrace -I -h {{inputs.parameters.radiance-parameters}} scene.oct < grid.pts > grid.res
outputs:
  artifacts:
  - name: result-file
    path: grid.res
```

### Recipe Path Context
When an artifact path is declared in a `Recipe DAG` it points to a folder or S3 bucket location where it expects the artifact to be pulled from (`input`) or pushed to (`output`). This path does not overwrite the one in the `Operator Function` context. Instead the executor (argo or luigi) will copy the artifact from the `Recipe Context` to the `Function Context` for `inputs` and vice versa for `outputs`

```yaml
name: main
inputs:
  parameters:
  - name: sensor-grid-count
    description: The maximum number of grid points per parallel execution
    default: 100
  - name: radiance-parameters
    description: The radiance parameters for ray tracing
    default: -I -ab 2 -h
  
  artifacts:
  - name: model
    description: A Honeybee model with radiance properties
    required: true
  - name: input-grid
    description: A grid file
    required: true

tasks:
   ...

- name: daylight-factor-simulation
  template: honeybee-radiance/ray-tracing
  dependencies:
  - split-grid
  - create-octree
  loop:
    from:
      type: tasks
      name: split-grid
      variable: grid-list
    sub_folder:
    - item.name
  arguments:
    parameters:
    - name: radiance-parameters
      from:
        type: inputs
        variable: radiance-parameters
    artifacts:
    - name: grid
      from:
        type: tasks
        name: split-grid
        variable: output-grids-folder
      subpath: '{{item.name}}'
    - name: scene-file
      from:
        type: tasks
        name: create-octree
        variable: scene-file
  outputs:
    artifacts:
    # This file is persisted in a loop sub-folder so the name won't clash
    - name: result-file
      path: grid.res
 ...
outputs:
  artifacts:
  - name: data
    from:
      type: tasks
      name: post-process
      variable: post-process-folder
  parameters:
  - name: average
    from:
      type: tasks
      name: post-process
      variable: daylight-factor-average
```

## Overhaul Documentation
The previous documentation page consisted of a Redoc template pointing to the old `Workflow` schema. As we have introduced new separate schemas as well as development practices for Operator and Recipe creation/management we needed to introduce some written tutorials and more in depth exploration pages for each schema. We therefore overhauled the Redoc page and went for a full Sphinx page that matches styling from other ladybug tools packages.


AntoineDao:
We have lift off!

PR #63, opened 2020-05-20
fix(recipe): don't lock dependencies when writing folder

AntoineDao:
Thanks for spotting this @mostaphaRoudsari !


AntoineDao:
I'll merge it in so you can keep working.

PR #64, opened 2020-05-20
Recipe review

mostaphaRoudsari:
This PR fixes two issues with recipe and dependency creation:

1. The URL handling for Windows is fixed.
2. The package URL is fixed.

@AntoineDao, the package_url was missing the subfolder. <s>Not sure if you like my implementation. I added it in the object when it returns the url but it can also be implemented in operator index.json itself to have the full url. `operators/energy-plus-0.1.0.tgz` instead of `energy-plus-0.1.0.tgz`.</s>

It was `http://localhost:8000/energy-plus-0.1.0.tgz` but should have been `http://localhost:8000/operators/energy-plus-0.1.0.tgz`

### DIFF
--- a/queenbee/base/io.py
+++ b/queenbee/base/io.py
@@ -34,7 +34,6 @@ class IOBase(BaseModel):
 
     artifacts: List[IOItem]
 
-    @classmethod
     @validator('parameters', 'artifacts')
     def parameter_unique_names(cls, v):
         """Pydantic validator to check that IO item names are unique within their list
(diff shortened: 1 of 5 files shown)


mostaphaRoudsari:
oops! it broke the tests. Let me check and see why the address is fine in the tests but not when I try to create a new recipe.

mostaphaRoudsari:
@AntoineDao, this is ready to be reviewed.

Issue #82, opened 2020-05-26
Local development experience


AntoineDao:
Had a chat with @mostaphaRoudsari yesterday and we agreed that the current developer experience for Queenbee is a bit slow. The main issue raised is that it is difficult to develop multiple Operators and Recipes at the same time because the following flow is required:

1. Update Operator defintion
2. Package and index (force refresh) modified operator
3. Refresh recipe dependencies (force overwrite dependency `hash`) 
4. ???
5. Profit

I think we might be able to use a similar mechanism to the [`npm link`](https://docs.npmjs.com/cli/link) command to ease this. Here is a good blog explaining how it works under the hood: [the magic behind NPM link](https://medium.com/@alexishevia/the-magic-behind-npm-link-d94dcb3a81af)

AntoineDao:
Let's call this resolved for now. We can re-open it or open another issue if more dev issues come up.

PR #84, opened 2020-05-26
Dev link

AntoineDao:
This PR aims to address #82. I more or less followed the system used by [`npm link`](https://docs.npmjs.com/cli/link) to achieve the following:

1. Develop an opertator `local-op` in `/path/to/operator/local-op`
2. Develop a recipe `local-rec` that has `local-op` as a dependency
3. Create a link between `local-op` and the dependencies folder of `local-rec` at `local-rec/.dependencies/operators/local-op`
4. Develop `local-op` and `local-rec` in tandem and create baked recipes from the `local-rec` folder

Let's show an example scenario where `local-rec` dependencies look like this:

```yaml
dependencies:
- type: operator
  name: honeybee-radiance
  hash: 4e6b9354ade52b792988abef4426d744b587fb26eaabbc681ad27c75aa0c8048
  version: 1.2.3
  source: https://example.com/test-repo
- type: operator
  name: local-op
  hash: 4e6b9354ade52b792988abef4426d744b587fb26eaabbc681ad27c75aa0c8048
  version: 1.2.3
  source: https://example.com/test-repo
```

The CLI command works like this:

```console
> cd local-op
> queenbee recipe link local-op /path/to/operator/local-op
```

In this command `local-op` is the name or alias of the dependency in the `dependencies.yml` file and `/path/to/operator/local-op` is the absolute or relative (it handles both) path to the operator folder we want to associate with this dependency in our `.dependencies` folder.

The link command will then create a `symlink` between `/path/to/operator/local-op` and `/path/to/recipe/local-rec/.dependencies/operator/local-op`.

**Warning**: Compiling a recipe into a baked recipe will only work if you compile it using the `from_folder` function and set `refresh_deps` to `False`. If you don't set `refresh_deps` to `False` queenbee will try to download the dependency from the defined source, check the `digest` hash etc...

### DIFF
--- a/tests/recipe/baked_recipe_test.py
+++ b/tests/recipe/baked_recipe_test.py
@@ -55,9 +55,12 @@ class TestFolder(BaseTestClass):
     def test_from_folder(self, folder):
         parsed_instance = BakedRecipe.from_folder(folder)
 
-        parsed_instance.to_yaml(
-            os.path.join(ASSET_FOLDER, 'baked', f'{parsed_instance.metadata.name}.yaml')
-        )
+        parsed_test_path = os.path.join(ASSET_FOLDER, 'baked', f'{parsed_instance.metadata.name}.yaml')
+
+        if os.path.exists(parsed_test_path):
+            compare_instance = BakedRecipe.from_file(parsed_test_path)
+
+            assert compare_instance == parsed_instance
 
     def test_from_recipe_instance(self, recipe):
         parsed_instance = BakedRecipe.from_recipe(recipe)
--- a/queenbee/cli/recipe/__init__.py
+++ b/queenbee/cli/recipe/__init__.py
@@ -2,6 +2,7 @@ from .new import new
 from .lint import lint
 from .package import package
 from .install import install
+from .link import link
 
 try:
     import click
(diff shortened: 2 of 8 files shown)


AntoineDao:
Before we merge this it would be awesome if we could test this on Windows. Not sure how `symlinks` work for windows...

tests/recipe/baked_recipe_test.py
```diff
@@ -55,9 +55,12 @@ def recipe_instances(self):
     def test_from_folder(self, folder):
         parsed_instance = BakedRecipe.from_folder(folder)
 
-        parsed_instance.to_yaml(
-            os.path.join(ASSET_FOLDER, 'baked', f'{parsed_instance.metadata.name}.yaml')
-        )
+        parsed_test_path = os.path.join(ASSET_FOLDER, 'baked', f'{parsed_instance.metadata.name}.yaml')
+
+        if os.path.exists(parsed_test_path):
```
mostaphaRoudsari, line 60: Shouldn't this be an assertion?

```suggestion
        assert os.path.exists(parsed_test_path)
```

AntoineDao: Nah, being able to read it is one thing. Checking whether it compiles correctly means we also have to write the compiled test which is work :smile:

PR #97, opened 2020-06-08
Auth

### LINKED ISSUE #93: Set up optional authentication config for dependency fetching
When `baking` a recipe we fetch all of it's dependencies from the specified source urls. Currently there is no way to make authenticated requests to those sources.

This means that we can't fetch dependencies from private repositories (ie: pollination cloud).

Here is a list of auth methods required in order of importance:
- [ ] JWT auth
- [ ] API token auth (maybe)
- [ ] Username/Password auth (maybe maybe...)

The Helm ChartMuseum [client auth](https://github.com/helm/chartmuseum#bearertoken-auth) system can be used as inspiration in conjunction with the Helm Auth Server [flow](https://github.com/chartmuseum/auth-server-example#helm-3).

### LINKED ISSUE #95: Change `version` to `tag`
If we are thinking of using versions like `latest` and other non semantically formatted version strings I think we should change the name of the `metadata.version` key to `metadata.tag` like Docker.

Any thoughts?

@cdriesler tagging you in this because I can't assign you.

### LINKED ISSUE #96: Generate tarfiles without a filesystem
It will be helpful when packaging recipes and operators in other contexts than a user's machine to be able to generate a tarfile bytes/file object without writing any data to files.

This is pretty crucial to be able to package operators/recipes on Pollination servers.

AntoineDao:
This PR addresses generating packaged tar files without using a filesystem (for server side packaging) and optional authentication (from library and CLI).

fixes #93
fixes #95
fixes #96

### DIFF
--- a/queenbee/cli/config/__init__.py
+++ b/queenbee/cli/config/__init__.py
@@ -12,10 +12,7 @@ except ImportError:
 @click.group('config')
 @click.pass_context
 def main(ctx):
-    """manage queenbee configuration data
-    
-
-    """
+    """manage queenbee configuration data"""
     pass
 
 main.add_command(view)
--- a/queenbee/cli/config/auth.py
+++ b/queenbee/cli/config/auth.py
@@ -10,18 +10,12 @@ except ImportError:
 
 @click.group('auth')
 def auth():
-    """manage queenbee configuration auth
-    
-
-    """
+    """manage queenbee configuration auth"""
     pass
 
 @auth.group('add')
 def add():
-    """add auth domains and methods to your queenbee config
-    
-
-    """
+    """add auth domains and methods to your queenbee config"""
     pass
 
 @add.command('pollination')
--- a/queenbee/config/auth.py
+++ b/queenbee/config/auth.py
@@ -32,6 +32,7 @@ class BaseAuth(BaseModel):
     def refresh_token(self):
         pass
 
+
 class JWTAuth(BaseAuth):
 
     type: Enum('JWTAuth', {'type': 'jwt'}) = 'jwt'
--- a/queenbee/repository/package.py
+++ b/queenbee/repository/package.py
@@ -211,7 +211,7 @@ class ResourceVersion(BaseModel):
 
         tar_file = TarFile.open(file_path)
 
-        version, _, _, _, _ = cls.unpack_tar(tar_file=filebytes, verify_digest=False)
+        version, *_ = cls.unpack_tar(tar_file=filebytes, verify_digest=False)
 
         return version
 


AntoineDao:
Also managed to sneak in some code to replace all `version` keys with `tag` to solve #95

queenbee/cli/config/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    )
+
+
+@click.group('config')
+@click.pass_context
+def main(ctx):
+    """manage queenbee configuration data
+    
+
+    """
```
mostaphaRoudsari, line 18: Are you trying to test my PEP8 flexibility? 😀

```suggestion
    """manage queenbee configuration data."""
```

queenbee/cli/config/auth.py
```diff
@@ -0,0 +1,48 @@
...
+
+@auth.group('add')
+def add():
+    """add auth domains and methods to your queenbee config
+    
+
+    """
+    pass
+
+@add.command('pollination')
```
mostaphaRoudsari, line 27: I believe this should be added from queenbee-pollination not in pollination itself. Otherwise we will have to add several plugin commands to the core library.

queenbee/cli/operator/package.py
```diff
@@ -32,19 +32,31 @@ def package(path, destination, force):
...
+    except ValueError as error:
+        raise click.ClickException(error)
+
+    file_path = os.path.join(destination, operator_version.url)
+
+    if not force and os.path.isfile(file_path):
+        raise click.ClickException(f'File already exists at path {file_path}. Use -f to overwrite it.')
+
+    with open(file_path, 'wb') as f:
+        file_object.seek(0)
```
mostaphaRoudsari, line 59: I'm not sure why would you need this line?

queenbee/config/auth.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @property
+    def auth_header(self):
+        if self.access_token is None:
+            return ''
+        return f'Bearer {self.access_token.get_secret_value()}'
+
+    def refresh_token(self):
+        pass
+
```
mostaphaRoudsari, line 34: 2 lines!

queenbee/repository/package.py
```diff
@@ -149,28 +206,40 @@ def from_package(cls, package_path: str):
         """
         file_path = os.path.normpath(os.path.abspath(package_path))
 
+        with open(file_path, 'rb') as f:
+            filebytes = BytesIO(f.read())
+
         tar_file = TarFile.open(file_path)
 
-        member = tar_file.getmember('version.json')
+        version, _, _, _, _ = cls.unpack_tar(tar_file=filebytes, verify_digest=False)
```
mostaphaRoudsari, line 214: This is cleaner:

```suggestion
        version, *_= cls.unpack_tar(tar_file=filebytes, verify_digest=False)
```

AntoineDao:
I'm in agreement about moving pollination stuff into plugins in general. In this case however I'm not 100% sure of what the best way to integrate special Auth requirements into non-plugin commands. In this case it felt acceptable much like the way Docker provides default auth methods for Dockerhub.

queenbee/cli/operator/package.py
AntoineDao: When we generate the package the output file object is a BytesIO stream. This stream is already "read" so has to be reset. It feels safer to do this before reading rather than assuming no other process will have consumed it.

queenbee/config/auth.py
AntoineDao: ?

queenbee/repository/package.py
AntoineDao: Nice! Will implement!

queenbee/cli/config/auth.py
AntoineDao: This is just a command to add "pollination" style Auth to the queenbee config file. It will not serve as an entry point for other pollination commands.

queenbee/config/auth.py
mostaphaRoudsari: You need two empty lines before the class. There was only one. PEP8 🚨

queenbee/cli/config/auth.py
mostaphaRoudsari: I understand that this is not the entry point command but don't you agree that this is a plugin specific command?

What stops you to move this command to queenbee-pollination? You can import `add` and then add your new command when queenbee-pollination is loaded. Right?

AntoineDao: Yeah it does kinda feel like a plugin command. The difficulty I'm having is not with integrating the click/cli side as a plugin but managing passing the `config` object which contains an Pollination api token and can create a JWT token that is then used to authenticate certain requests within recipes.

Essentially if I make this command a plugin then it feels I have to make this [class](https://github.com/ladybug-tools/queenbee/pull/97/commits/84add41f8789f2188eae41a36d8b72f747456f01#diff-22a2a60dc5ac2d09ddd52a6343eae14cR35-R93) which then makes it difficult (I think) to handle authentication to pollination servers when managing recipe dependencies. Part of the reason this is difficult is because we don't (currently) implement standard authentication mechanisms which we can expect to be reusable by other non-pollination servers.

queenbee/cli/config/__init__.py
AntoineDao: My bad... I... have no excuses at this point...

AntoineDao:
@mostaphaRoudsari I am going to merge this after having fixed the styling issues. I am in agreement with you about plugin based loading of pollination specific commands and modules/functions however it will take me time to think about and implement a decent plugin system that I won't yell at and rip apart in 2 weeks time.

I have raised #98 to track this issue and we can plan some work to remove pollination specific modules and commands (it's only 1 class and one command for now) from queenbee.

PR #114, opened 2020-07-19
fix(artifacts_source): replace Enum with constr in type

mostaphaRoudsari:
This was causing an issue with the documentation: https://travis-ci.org/github/ladybug-tools/queenbee/jobs/708701110#L536

I believe this is a result of the recent change in how Pydantic works with Enum objects. We can use the Enum trick anymore. The good news is that `constr` is there to help.

cc: @AntoineDao, @chriswmackey

### DIFF
--- a/queenbee/recipe/artifact_source.py
+++ b/queenbee/recipe/artifact_source.py
@@ -4,8 +4,7 @@ ArtifactSource is a configuration to a source system to acquire artifacts from.
 
 """
 from typing import Dict, List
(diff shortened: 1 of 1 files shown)


AntoineDao:
Using `constr` does partially solve this but it results in non-explicit JSON schema and openapi definitions for these objects. It's not a big deal but I would be keen to use enum where it makes sense to (after fixing for the pydantic issue)

PR #125, opened 2020-08-11
fix(task): add support for fixed value in task artifact arguments

### LINKED ISSUE #121: Add an option to hard-code input values for DAGTaskArguments and artifacts
We currently only support inputs that are referenced in DAG's input or a previous task's output (https://github.com/ladybug-tools/queenbee/blob/cd316ea4feeea87410e8aa9c5f581822ea76d8db/queenbee/recipe/dag.py#L203-L205) but there is another case when the input needs to be hard-coded in task level itself.

Two examples that are currently blocking are aggregated results folder and gensky with fixed value for daylight factor.

mostaphaRoudsari:
Added support for fixed value for artifact argument for a task. This is similar to what a parameter argument supports and is the minimum change that I need to be able to get the daylight-factor results aggregate work.

We need to add support for this in queenbee extensions.

Resolves #121

### DIFF
--- a/queenbee/recipe/dag.py
+++ b/queenbee/recipe/dag.py
@@ -187,6 +187,18 @@ class DAGTaskArtifactArgument(BaseModel):
         ' want to source an artifact from within that directory.'
     )
 
+    value: str = Field(
+        None,
+        description='The fixed value for this task argument.'
+    )
+
+    @validator('value')
+    def check_value_exists(cls, v):
+        if v is not None:
+            assert v.from_ is not None, \
+                ValueError('value must be specified if no "from" source is specified for argument artifact.')
+        return v
+
 
 class DAGTaskParameterArgument(BaseModel):
     """Input argument for a DAG task.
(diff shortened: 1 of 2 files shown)


queenbee/recipe/dag.py
```diff
@@ -187,6 +187,18 @@ class DAGTaskArtifactArgument(BaseModel):
         ' want to source an artifact from within that directory.'
     )
 
+    value: str = Field(
```
AntoineDao, line 190: Not sure I understand what the `fixed value` is here. Is it a JSON object? Is it a hardcoded path to something in the project folder?

If it's the latter then I think we should create a `ProjectFolderPathReference` class to put in the `from_` union.

AntoineDao:
@mostaphaRoudsari check out my proposed change and let me know what you think. (I can't set you as reviewer to your own PR unfortunately...)

The proposed change means that that task can no look like this:

```yaml
name: some-task
template: honeybee-energy/simulate
arguments:
  artifacts:
  - name: fixed-file-path
    from:
      type: folder
      path: path/to/file/in/workflow/folder
```

mostaphaRoudsari:
Hi @AntoineDao, thanks. This looks fine to me. I tried to replicate the pattern that we had for parameter argument but this can also totally work. Just need a way to set it up to a fixed value! 😃

PR #132, opened 2020-08-24
fix(dag): fix validation for arguments value

mostaphaRoudsari:
I found this bug while updating daylight recipes. Once merged we can push the new recipe.

### DIFF
--- a/queenbee/recipe/dag.py
+++ b/queenbee/recipe/dag.py
@@ -230,7 +230,7 @@ class DAGTaskParameterArgument(BaseModel):
     @validator('value')
     def check_value_exists(cls, v, values):
         if v is None:
-            assert values['from_'] is not None, \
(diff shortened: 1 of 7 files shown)


AntoineDao:
The general rule of thumb for validation bugs that we pick up is that we should write tests to confirm the behaviour and then fix the code to work with the given test.

This should ensure that the new code works as expected and is backwards compatible.

In the case of 93b1164 the reason they failed was because some of the test files were incorrect 😅 I fixed the broken test files in 652c9a1.

AntoineDao (approved):

AntoineDao:
PS: again, sorry for pushing straight to your branch. Feel free to change what you don't agree with and push again. It's just easier to communicate and propose change on the same branch for PRs 😄

Issue #143, opened 2020-09-21
Match correct type to one of AnyOf


MingboPeng:
There are multiple places have invalid types for AnyOf type parameter. 

For example:
In DAGTaskArtifactArgument, parameter "from" requires the value has to be one of `InputArtifactReference` , `TaskArtifactReference` ,`FolderArtifactReference`, 
![image](https://user-images.githubusercontent.com/9031066/93825707-ce20a900-fc33-11ea-9852-e6d7ea8866e8.png)

but the task inside `annual daylight` recipe contains "folder" type which doesn't match any of above three required types.

![image](https://user-images.githubusercontent.com/9031066/93825479-5f435000-fc33-11ea-9014-bf583322ac1d.png)

We need to validate the recipes before publishing them.

AntoineDao:
The `type` value for the artifact in your example is correct according to the [OpenAPI schema](https://www.ladybug.tools/queenbee/_static/redoc-recipe.html#tag/dagtaskartifactargument_model):
![image](https://user-images.githubusercontent.com/20436635/93853540-edf3b380-fcab-11ea-8a75-f445d4354d3e.png)

The `FolderArtifactReference` has two properties `type` and `path` where `type` is an `enum` value which can only be `folder`.

What issue or bug have you encountered that made you conclude that the `type` strings are incorrect?

MingboPeng:
Hi @AntoineDao,

This is issue with Union(AnyOf) type.
In AnyOf type, in this case: FolderArtifactReference is one of valid type. However, from plain json perspective, following text cannot be de-serialized back to FolderArtifactReference object since the Json Serializer has no way to know which type is correct type amount InputArtifactReference, TaskArtifactReference, FolderArtifactReference. 
```
{
"type": "folder",
"path": "initial_result/{{item}}/final"
}
```

The way how HoneybeeSchema does is that ensure "type" value is the type name that Json Serializer can use for de-serialization. 

One side question here: how does this current enum type helpful if there is only one value in it? is it only for checking typos instead of just using string?

PR #144, opened 2020-09-21
fix(reference): set type to parameter/artifact Reference

MingboPeng:


MingboPeng:
Addressing https://github.com/ladybug-tools/queenbee/issues/143

AntoineDao (changes requested):
We can switch to using `constr` to avoid having an `Enum` value for types if it makes it easier for dotnet 😄

I would rather not change the string value for the types though as this introduces breaking changes to the schema. 

That being said, could you explain why you think we need the `type` name to match the class name? Is this the way you have designed the honeybee schema? Was it for dotnet SDK purposes or was there another reason behind this?

MingboPeng:
Thanks @AntoineDao, I try to explain it here https://github.com/ladybug-tools/queenbee/issues/143#issuecomment-696747587.

mostaphaRoudsari:
This will be a breaking change and we will address it with the other breaking changes to Queenbee!

mostaphaRoudsari:
I'm closing this PR in favor of #152 which will address this issue.

PR #148, opened 2020-09-27
Fixes in order to get the local indexing to work

mostaphaRoudsari:
There are two commits that are related to development issues here:

- One that fixes the path issues for Windows path
- A second one that fixes the path for accessing the local resources

With these fixes I could create a local index and get `queenbee repo list` to work. Once this is merged I need to:

1. Send a new PR to honeybee-radiance-recipe repository with the new file structure.
2. Add a new command to Queenbee luigi to support the new way of getting the recipe from a repository instead of a baked recipe file.
3. Use the new command the Grasshopper component to deal with honeybee-radiance recipes.

### DIFF
--- a/queenbee/base/request.py
+++ b/queenbee/base/request.py
@@ -12,7 +12,10 @@ def get_uri(url):
 
     if url.startswith('file:///'):
         return url.replace('\\', '/')
+    elif url.startswith('http://') or url.startswith('https://'):
+        return url
 
+    # a local file 
     try:
         uri = pathlib.Path(os.path.abspath(url)).as_uri()
     except ValueError:
--- a/queenbee/base/request.py
+++ b/queenbee/base/request.py
@@ -1,9 +1,24 @@
+import pathlib
+import os
 from urllib import request
 from typing import Union
-
 from .basemodel import BaseModel
 
 USER_AGENT_STRING = 'Queenbee'
+
+
+def get_uri(url):
+    """Resolve uri for urls and local files."""
+
+    if url.startswith('file:///'):
+        return url.replace('\\', '/')
+
+    try:
+        uri = pathlib.Path(os.path.abspath(url)).as_uri()
+    except ValueError:
+        uri = url
+
+    return uri
 
 
 def urljoin(*args):
@@ -13,7 +28,7 @@ def urljoin(*args):
         str: a clean url string
     """
     url = "/".join(map(lambda x: str(x).rstrip('/'), args))
-    url.replace('\\', '/')
+    url = url.replace('\\', '/')
 
     return url
 
--- a/queenbee/cli/repository/init.py
+++ b/queenbee/cli/repository/init.py
@@ -20,7 +20,7 @@ def init(path):
 
     path = os.path.abspath(path)
 
-    os.makedirs(os.path.join(path, 'workflows'), exist_ok=True)
+    os.makedirs(os.path.join(path, 'operators'), exist_ok=True)
     os.makedirs(os.path.join(path, 'recipes'), exist_ok=True)
 
     index = RepositoryIndex.from_folder(path)
--- a/queenbee/cli/repository/manage.py
+++ b/queenbee/cli/repository/manage.py
@@ -1,6 +1,5 @@
 import os
 import json
-from urllib.parse import urlparse
 
 try:
     import click
@@ -25,12 +24,6 @@ def add(name, path, force):
     is useful to search, develop and use queenbee packages locally.
     """
     ctx = click.get_current_context()
-
-    parse_res = urlparse(path)
-
-    # Not a url, therefore local path
-    if parse_res.scheme == '':
-        path = os.path.abspath(path)
 
     try:
         ctx.obj.config.add_repository(
--- a/queenbee/config/repositories.py
+++ b/queenbee/config/repositories.py
@@ -31,7 +31,7 @@ class RepositoryReference(BaseModel):
         from ..repository import RepositoryIndex
 
         if self.path.startswith('file:'):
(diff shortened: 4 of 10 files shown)


queenbee/base/request.py
```diff
@@ -1,19 +1,34 @@
...
 
 
+def get_uri(url):
+    """Resolve uri for urls and local files."""
+
+    if url.startswith('file:///'):
+        return url.replace('\\', '/')
+
+    try:
+        uri = pathlib.Path(os.path.abspath(url)).as_uri()
```
AntoineDao, line 17: Why are we trying to resolve an http url as an absolute path? Lines 13-14 address the case where a url is for a local path so maybe we should make that one an `abspath`.

What happens if we keep this line is that all the `http` urls are appended with a local file path like so:

```
file:///root/projects/ladybug-tools/queenbee/https%3A/api.pollination.cloud/registries/antoinedao
```

queenbee/base/request.py
```diff
@@ -1,19 +1,34 @@
...
+
 def urljoin(*args):
     """join multiple string into a clean url
 
     Returns:
         str: a clean url string
     """
     url = "/".join(map(lambda x: str(x).rstrip('/'), args))
-    url.replace('\\', '/')
+    url = url.replace('\\', '/')
```
AntoineDao, line 31: Good catch... 😅

queenbee/cli/repository/init.py
```diff
@@ -20,7 +20,7 @@ def init(path):
 
     path = os.path.abspath(path)
 
-    os.makedirs(os.path.join(path, 'workflows'), exist_ok=True)
+    os.makedirs(os.path.join(path, 'operators'), exist_ok=True)
```
AntoineDao, line 23: Good catch... 😅

queenbee/cli/repository/manage.py
```diff
@@ -26,12 +25,6 @@ def add(name, path, force):
     """
     ctx = click.get_current_context()
 
-    parse_res = urlparse(path)
```
AntoineDao, line 29: Thanks for dedupelicating this 😀

queenbee/config/repositories.py
```diff
@@ -37,9 +31,7 @@ def fetch(self, auth_header: str = '') -> 'RepositoryIndex':
         from ..repository import RepositoryIndex
 
         if self.path.startswith('file:'):
-            path = self.path.split('file:')[-1]
-            index_path = os.path.join(path, 'index.json')
-            url = f'file:{index_path}'
+            url = os.path.join(self.path, 'index.json').replace('\\', '/')
```
AntoineDao, line 34: If we are already sanitising the path input when validating the `RepositoryReference.path` value do we still need to use `replace('\\', '/')`?

queenbee/repository/package.py
```diff
@@ -274,12 +274,12 @@ def from_package(cls, package_path: str):
 
     def fetch_package(self, source_url: str = None, verify_digest: bool = True, auth_header: str = '') -> 'PackageVersion':
         if source_url.startswith('file:'):
-            source_path = source_url.split('file:')[1]
-
+            source_path = source_url.split('file:///')[1]
+            subfolder = f'{self.type}s'
             if os.path.isabs(source_path):
-                package_path = os.path.join(source_path, self.url)
+                package_path = os.path.join(source_path, subfolder, self.url)
```
AntoineDao, line 280: We don't need to add the `subfolder` here. The `self.url` is be a valid relative path from the `source_path`.

image

AntoineDao (changes requested):
Thanks for the catches and for refactoring the path management functions. Two things are breaking when I run the code locally so we should change them or discuss them for me to better understand what you're after:

1. Trying to create an absolute path from an `http` url in `queenbee/base/request.py` creates a double path which is invalid
2. There should be no need to add a `subfolder` value to a `package_path` when fetching from a repository because the package `url` should be a valid relative path from the repository `index.json`

queenbee/base/request.py
mostaphaRoudsari: Good catch! I will fix this.

queenbee/config/repositories.py
mostaphaRoudsari: You are right. This one is redundant knowing the the path is already cleaned. Let me remove this one.

queenbee/repository/package.py
mostaphaRoudsari: This is very different from what I get for a local repository:

image

We need the subfolder to get the local ones to work.

mostaphaRoudsari: This commit should fix the bug for adding subfolder name to the url.

https://github.com/ladybug-tools/queenbee/pull/148/commits/e4928355a29aafa2832f8b0b6b961dac0d52e083

mostaphaRoudsari:
@AntoineDao, both of your comments should be addressed. Let me know if you have any other comments.

queenbee/base/request.py
AntoineDao: Still not sure this is resolved. From my understanding a `url` should only start with either: `file://`, `http://` or `https://`. Therefore lines 18/24 just look like a no-op to me.

AntoineDao (approved):
Still not 100% sure what's happening in `queenbee/base/request.py` as per the comment I unresolved. Other than that it's looking good! 👍

queenbee/base/request.py
mostaphaRoudsari: You hit line 18-24 the first time that you add a local repository. In that case `url` is a file but doesn't start with a `file:///`.

`queenbee repo add honeybee-radiance usr/path/to/local/repo`

AntoineDao: Oh! Great catch! 😅

PR #152, opened 2020-10-02
Queenbee refactor - input / outputs

mostaphaRoudsari:
This commit takes out the inputs and output from dag module and put them under an IO module. The submodules are broken down into dag, function and task.

Here is the list of most most important changes:

- DAG and Operator inputs are now very similar objects.
- The separation between parameters and artifacts doesn't exist anymore. There is only a list of inputs and a list of outputs.
 - Inputs can be from either of these types:
    - StringInput
    - IntegerInput
    - NumberInput
    - BooleanInput
    - FolderInput
    - FileInput
    - PathInput
    - ArrayInput and ObjectInput are also implemented

- Every input and output now has a type field
  
  ```
  - name: model
    type: DAGFolderInput
    description: Path to Radiance model folder.
    path: model
  ```

- Folder, File and Path inputs represent what used to be artifacts
- All the inputs now have a `spec` field for adding additional JSONSchema specification. This will help us to standardize the input validation among all the UIs. Since JSONSchema validators are available in all the major programming languages each UI can do the validation before submitting the recipe.
  ```
  - name: include-aperture
    type: StringInput
    description: |
      A value to indicate if the static aperture should be included in octree. Valid
      values are include and exclude. Default is include.
    default: include
    spec:
      type: string
      enum:
      - include
      - exclude
  ```

- The required field has been changed to a property. Input.required returns true if there is no default value and False if there is one.

## BREAKING CHANGES

By removing artifacts and parameters distinction the older Queenbeee schema is no longer compatible.


queenbee/io/__init__.py
```diff
@@ -30,9 +30,9 @@ class IOBase(BaseModel):
     IOBase is used within Operators, Recipes and Workflows.
     """
 
-    parameters: List[IOItem]
+    inputs: List[IOItem]
```
MingboPeng, line 33: I like this List[IOItem] instead of List[Union[xxx,xxx,xxx,xxx,xxx,xxx,xxx,xxx]]

queenbee/io/common.py
```diff
@@ -0,0 +1,133 @@
...
+
+class ItemType(str, Enum):
+    """Type enum for items in a list."""
+    String = 'string'
+    Integer = 'integer'
+    Number = 'number'
+    Boolean = 'boolean'
+    Folder = 'folder'
+    Array = 'array'
+    Object = 'object'
```
MingboPeng, line 20: These reserved keywords works fine with Enum in C# with CamelCase.
image

But not sure if this order works:
String = 'string'  (CamelCase = 'lowercase')
 
In [Honeybee schema, it is flipped](https://github.com/ladybug-tools/honeybee-schema/blob/071e5e3d5200a435f9970e4236ff6c892d96e0dd/honeybee_schema/energy/designday.py#L156-L167):
image

queenbee/io/common.py
```diff
@@ -0,0 +1,133 @@
...
+        ...,
+        description='Input name.'
+    )
+
+    description: str = Field(
+        None,
+        description='Optional description for input.'
+    )
+
+    default: Any = Field(
```
MingboPeng, line 38: Can this "Any" changed to object?

mostaphaRoudsari: This is a placeholder and will be overwritten by every IO type. You should never get an input with such default but I can change it to String if that is better.

queenbee/io/common.py
mostaphaRoudsari: Thanks. Will make the change.

queenbee/io/__init__.py
mostaphaRoudsari: We can't do that anymore! This code will also change! 😀

This is now similar to Radiance materials. We have to handle all the different types. JSONSchema/OpenAPI doesn't support check for subclasses. It should be the exact object type.

queenbee/io/function.py
```diff
@@ -0,0 +1,222 @@
...
+        if self.extensions:
+            assert value.lower().endswith(self.extension.lower()), \
+                f'Input file extension for {value} must be {self.extensions}'
+        if self.spec:
+            spec = dict(self.spec)
+            spec['type'] = 'string'
+            json_schema_validator(value, spec)
+
+
+class FunctionPathInput(FunctionFileInput):
```
MingboPeng, line 149: The current hierarchy is **FunctionFolderInput** => **FunctionFileInput** => **FunctionPathInput**.

**FunctionPathInput** doesn need the "extension" inherited from **FunctionFileInput**, it'd be better to have:
**FunctionPathInput** => **FunctionFolderInput** 
**FunctionPathInput** => **FunctionFileInput**

queenbee/io/reference.py
```diff
@@ -0,0 +1,256 @@
...
+
+    @property
+    def source(self):
+        return None
+
+    def fetch_from_flow(self, flow):
+        pass
+
+
+class FileReference(_BaseReference):
```
MingboPeng, line 35: Are these "Reference" only used inside the tasks? Won't be used as a input type for recipe, right?

queenbee/workflow/arguments.py
```diff
@@ -15,7 +15,7 @@ class ArgumentArtifact(BaseModel):
         description='The name of the artifact'
     )
 
-    source: Union[HTTPSource, S3Source, ProjectFolderSource] = Field(
+    source: Union[HTTP, S3, ProjectFolder] = Field(
```
MingboPeng, line 18: Can these three HTTP, S3, ProjectFolder types merged into one type say "ArgumentArtifactSource" and use enum for HTTP, S3, ProjectFolder.

```
class ArgumentArtifactSourceTypes(str, Enum):
    HTTP = 'HTTP'
    S3= 'S3'
    ProjectFolder= 'ProjectFolder'

class ArgumentArtifactSource(_ArtifactSource):
    source_type: ArgumentArtifactSourceTypes
    path: str = Field( ................ )
```

mostaphaRoudsari: I don't think so because each of them are a different object with information and not a string.

queenbee/io/reference.py
mostaphaRoudsari: Yes. These are just reference objects which are similar to wires in Grasshopper. It says from where should I source the values for an input.

queenbee/io/function.py
mostaphaRoudsari: Sure. Will have a look and change this

mostaphaRoudsari: I read this again. FunctionPath actually need the extension in case the input is a file input. Path can be a file or a folder.

queenbee/recipe/dag.py
```diff
@@ -381,31 +22,12 @@ class DAGTaskLoop(BaseModel):
     input or task parameter which should be a list.
     """
 
-    from_: Union[InputParameterReference, TaskParameterReference] = Field(
+    from_: Union[InputReference, TaskReference, ValueListReference] = Field(
```
AntoineDao, line 25: Am I right in thinking that the three references types all represent `parameters` and not `artifacts`?

queenbee/recipe/dag.py
```diff
@@ -420,14 +42,15 @@ class DAGTask(BaseModel):
...
 
-    arguments: DAGTaskArgument = Field(
+    arguments: List[TaskArguments] = Field(
         None,
-        description='The input arguments for this task'
+        description='The input arguments for this task.'
     )
 
-    dependencies: List[str] = Field(
+    needs: List[str] = Field(
```
AntoineDao, line 50: This will make people happy 😀

queenbee/recipe/dag.py
```diff
@@ -442,22 +65,22 @@ class DAGTask(BaseModel):
         'wanting to save results in a specific folder.'
     )
 
-    outputs: DAGTaskOutputs = Field(
+    returns: List[TaskReturns] = Field(
```
AntoineDao, line 68: This will also make things less confusing hopefully 😀

queenbee/workflow/arguments.py
AntoineDao: I agree with @mostaphaRoudsari  on this one. The `HTTP`, `S3` and `ProjectFolder` classes have more properties than just `path` so they are different objects.

```yaml
type: ProjectFolder
path: C:\Users\Pollination\my-model.json
---
type: HTTP
url: https://models.pollination.cloud/my-model.json
---
type: S3
key: path/to/model.json
endpoint: https://s3.pollination.cloud
bucket: my-project
```

AntoineDao (commented):
Overall this is looking really good! I'm enjoying using the `io` folder for all things "types and validation". I left a few comments/questions in the code and have two broader questions for you regarding this PR.

### Annotations on every object
I am not sure we want to provide capacity annotate every single object. I would preserve it for `functions`, `tasks`, `dags` and `recipes`. I think we can remove `annotations` from the IO Type classes in favour of the enhanced typing system they provide. So if someone on an UI wants to know something about an input or output they will rely on the type schemas rather than annotations.

I am referring to [this line of code](https://github.com/ladybug-tools/queenbee/pull/152/files#diff-59ebb7751ffcd7b74aa39ed01756c317R30) where `annotations` are added to the `BaseModel` class that is inherited by pretty much everything in Queenbee.

### Types for validation and types for UI building
Am I right in thinking that the Task Argument, Task Returns and DAG Output `type`  values are mostly there to facilitate building a UI rather than perform validation? 

I assume we would still check:
1. The Task Argument Reference is valid against the input of the Function or DAG it refers to
2. The DAG Output Reference is determined and validated by fetching the output type of the Task Return object, which itself checks the output of the Function or DAG (recursively) it refers to


PS: Can't approve in good conscience because the tests are failing 😝

queenbee/io/common.py
mostaphaRoudsari: @MingboPeng, do you think it will be a problem in the baseclass? I can remove it and copy the validator to all the subclasses. There will be some duplicated code but other than that it should be fine.

queenbee/io/reference.py
mostaphaRoudsari: Here is a supporting graphic.

inputs_outpus

queenbee/io/common.py
mostaphaRoudsari: The change has been made.

queenbee/base/basemodel.py
```diff
@@ -26,6 +27,17 @@ class BaseModel(PydanticBaseModel):
     """BaseModel with functionality to return the object as a yaml string
 
     """
+    annotations: Dict[str, str] = Field(
+        None,
+        description='An optional dictionary to add annotations to inputs. These '
+        'annotations will be used by client side libraries to validate the inputs or '
+        'bind them to specific actions. Use ``schema`` key for providing JSON Schema '
```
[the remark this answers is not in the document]
mostaphaRoudsari: Good catch! I fixed it.

queenbee/io/reference.py
```diff
@@ -0,0 +1,315 @@
...
+class _BaseReference(BaseModel):
+    """A Base reference model."""
+
+    type: constr(regex='^_BaseReference$') = '_BaseReference'
+
+    @property
+    def source(self):
+        return None
+
+    def fetch_from_flow(self, flow):
```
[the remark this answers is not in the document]
mostaphaRoudsari: Good question but I was thinking to ask that from you! 😁 It was in the original code and I just copy/pasted it. Since there was not enough information on what it does I didn't dare to delete it. Do you know if you use it in any shape or form?
If not we should just remove it.

queenbee/recipe/dag.py
mostaphaRoudsari: Yes. If you don't see Folder/File/Path in the name of the reference then it is referring to a parameter.

queenbee/recipe/dag.py
mostaphaRoudsari: I hope so. Since it is a breaking change. 😵

mostaphaRoudsari:
Hi @AntoineDao, Thank you for reviewing the code and providing feedback. 

> I think we can remove annotations from the IO Type classes in favour of the enhanced typing system they provide. So if someone on an UI wants to know something about an input or output they will rely on the type schemas rather than annotations.

The reasoning behind keeping the annotations for the IO Type was to support extra information on platform by platform basis. A good example is changing the Grasshopper input for model_json to the real python Model object or to the real model in Rhino. We were thinking to use the inputs annotations field for DAG. Another example is loading the results from outputs before exposing them in Grasshopper.

I'm going to improvise here for an example and I know we are not allowing using lists in annotations value but maybe we should!

```JSON
"name": "model_folder",
"type": "DAGFileInput",
"annotations":
  {
    "grasshopper/alias": "model",
    "grasshopper/type": "NoTypeHint",
    "grasshopper/description": "A Honeybee model object",
    "grasshopper/action/method": "to.radiance_folder",
    "grasshopper/action/args": ["self", "temp"],
    "rhino/action": "link"
  }
```

I agreed that it will not be useful for Task Arguments and TaskReturns but having them there will not really hurt. If you have strong feeling about this I can add it to specific objects and remove it form TaskArgument and TaskReturn.

> I assume we would still check [...]

Yes. That is the plan. I yet to write those validations of course.

queenbee/io/common.py
MingboPeng: I am not sure what this Any will be translated to in OpenApi. Let's keep it for now and get back to this if it is an issue

mostaphaRoudsari:
I don't know what happened to Travis but the test are running fine locally.

mostaphaRoudsari:
I'm closing this PR in favor of the new one which is to a different branch to keep master as-is for now.

PR #157, opened 2020-10-27
Add input and output types

mostaphaRoudsari:
See https://github.com/ladybug-tools/queenbee/pull/152 for why and how.


mostaphaRoudsari:
@AntoineDao and @MingboPeng, since we already had two rounds of review process about these changes I'm thinking to merge this i and send smaller PRs for the new changes so you can review those. Let me know if that's OK with you.

AntoineDao:
Mind letting me have a quick look over for some details? I notice you've added the "doc.py" for example which might clash with existing sphinx documentation mechanism. I'll have a review for you in the next 3 hours 😀

mostaphaRoudsari:
Sounds good. I'm not really using docs.py right now. It's a place holder to generate what we need on dotnet side. I can either remove it for now and add it back in an upcoming PR or open an issue to integrate it with sphinx documentation.

queenbee/base/variable.py
```diff
@@ -19,6 +16,7 @@ def get_ref_variable(value: Union[bytes, str]) -> list:
     return parse_double_quotes_vars(value)
 
 
+# TODO: update workflow inputs validation
 def _validate_workflow_var_format(value: str) -> str:
```
AntoineDao, line 20: This is used by the `validate_ref_variable_format` function below but I cannot see any part of the codebase where the `validate_ref_variable_format` function is used. Do we need it for queenbee-luigi? If not I think we can dump these functions entirely as I don't use them in queenbee-argo.

For reference: I think we only use the `{{inputs.some-key}}` form in function commands.

queenbee/cli/recipe/new.py
```diff
@@ -30,37 +30,37 @@ def new(name, path):
         \b
         metadata:
             name: <input-name>
-            version: 0.1.0
+            tag: 0.1.0
```
AntoineDao, line 33: Good catch thanks!

queenbee/cli/recipe/new.py
```diff
@@ -30,37 +30,37 @@ def new(name, path):
...
         tasks:
         - name: say-something
             template: whalesay/say-hi
             arguments:
-            parameters:
             - name: message
+                type: TaskArgument
                 from:
-                type: inputs
+                type: InputReference
```
AntoineDao, line 53: This looks like it needs to be indented (as does the line below).

docs.py
```diff
@@ -0,0 +1,89 @@
...
+    "license": {
+        "name": "MIT",
+        "url": "https://github.com/ladybug-tools/queenbee-schema/blob/master/LICENSE"
+    }
+}
+
+
+modules = [
+    {'module': Recipe, 'name': 'Recipe'}
+    # {'module': Operator, 'name': 'Operator'}
```
AntoineDao, line 49: Should we not also generate Operator and Workflow Schemas?

AntoineDao (approved):
Looking really good! I have made a couple of minor comments.

### OpenAPI Schema Generation
So this method won't clash with the one set up to generate the Sphinx docs. That being said it would be nice to integrate `docs.py` as part of the `Queenbee` module rather than a random script outside of it. That way we can leverage the schema generation in the docs as well as the SDK generation.

An additional thought on this note: have we considered trying the JSON schema of the Queenbee Pydantic classes rather than the OpenAPI schemas? They are slightly different and might result in better SDK generation in C#.

### Testing System
Now that you have had the joy of rewriting a bunch of tests I figured it's the right time to ask: what do you think of the testing system? Is it helpful to generate many validation cases? Or is it too much work to fix?

docs.py
mostaphaRoudsari: Good catch! Thanks.

queenbee/cli/recipe/new.py
mostaphaRoudsari: Good catch! Thanks.

queenbee/base/variable.py
mostaphaRoudsari: I don't use it too. Let's remove it.

docs.py
mostaphaRoudsari: https://github.com/ladybug-tools/queenbee/issues/158

mostaphaRoudsari:
Hi @AntoineDao, thank you for the review.

> That being said it would be nice to integrate docs.py as part of the Queenbee module rather than a random script outside of it. That way we can leverage the schema generation in the docs as well as the SDK generation.

Agreed! I opened an issue to address this soon.

> An additional thought on this note: have we considered trying the JSON schema of the Queenbee Pydantic classes rather than the OpenAPI schemas? They are slightly different and might result in better SDK generation in C#.

I don't know how the the tooling around JSON schema works. Since we are using OpenAPI gen and it is working I'm thinking to keep it as is.

> Now that you have had the joy of rewriting a bunch of tests I figured it's the right time to ask: what do you think of the testing system? Is it helpful to generate many validation cases? Or is it too much work to fix?

I think the way pytest is used is fine. I found updating the baked recipes from folder hard and to some extent not really useful since the way that I generate them is basically what the test is checking. Overall it is good that we have more tests.

PR #165, opened 2020-11-03
Add alias input and output objects

mostaphaRoudsari:
This PR adds alias object for inputs and outputs that allows the recipe author to assign pre/post processing for each input/output in different client side applications like Grasshopper, Rhino, etc.

After testing different ideas I ended up adding an alias input/output type for each input/output. This will allow the developers to create a new valid input/output for the target platform with no limitations. The handler will take care of converting the alias input to the original input.


mostaphaRoudsari:
wrong branch!

PR #166, opened 2020-11-03
Add alias input and output objects

mostaphaRoudsari:
This PR adds alias object for inputs and outputs that allows the recipe author to assign pre/post processing for each input/output in different client side applications like Grasshopper, Rhino, etc.

After testing different ideas I ended up adding an alias input/output type for each input/output. This will allow the developers to create a new valid input/output for the target platform with no limitations. The handler will take care of converting the alias input to the original input.

### DIFF
--- a/queenbee/io/alias.py
+++ b/queenbee/io/alias.py
@@ -1,4 +1,9 @@
-"""Queenbee input and output types for a DAG."""
+"""Queenbee alias input and output types.
+
+Use these alias inputs and outputs to create a different IO object for client side UIs.
+Alias inputs provide a handler object to convert the input provided to alias to what is
+required with the original input.
+"""
 
 import os
 from typing import Dict, Union, List
@@ -6,32 +11,77 @@ from typing import Dict, Union, List
 from pydantic import constr, Field, validator
 from jsonschema import validate as json_schema_validator
 
-from .common import ItemType, GenericInput, FromOutput
+from .common import ItemType, GenericInput, FromOutput, find_dup_items
 from .artifact_source import HTTP, S3, ProjectFolder
-from .alias import DAGAliasInputs, DAGAliasOutputs
 from ..io.reference import FileReference, FolderReference, TaskReference
+from ..base.basemodel import BaseModel
 
 
-class DAGGenericInput(GenericInput):
-    """Base class for DAG inputs.
+class IOAliasHandler(BaseModel):
+    """Input and output alias handler object."""
+
+    type: constr(regex='^IOAliasHandler$') = 'IOAliasHandler'
+
+    language: str = Field(
+        ...,
+        description='Declare the language (e.g. python, csharp, etc.). This '
+        'option allows the recipe to be flexible on handling different programming '
+        'languages.'
+    )
+
+    module: str = Field(
+        ...,
+        description='Target module or namespace to load the alias function.',
+        example='honeybee_rhino.handlers'
+    )
+
+    function: str = Field(
+        ...,
+        description='Name of the function. The input value will be passed to this '
+        'function as the first argument.'
+    )
+
+
+class DAGGenericInputAlias(GenericInput):
+    """Base class for DAG Alias inputs.
 
     This class adds a handler to input to handle the process of loading the input
     from different graphical interfaces.
     """
-    type: constr(regex='^DAGGenericInput$') = 'DAGGenericInput'
+    type: constr(regex='^DAGGenericInputAlias$') = 'DAGGenericInputAlias'
 
-    alias: List[DAGAliasInputs] = Field(
-        None,
-        description='A list of aliases for this input in different platforms.'
+    platform: List[str] = Field(
+        ...,
+        description='Name of the client platform (e.g. Grasshopper, Revit, etc). The '
+        'value can be any strings as long as it has been agreed between client-side '
+        'developer and author of the recipe.'
     )
 
-    @validator('alias', always=True)
-    def create_empty_handler_list(cls, v):
+    handler: List[IOAliasHandler] = Field(
+        ...,
+        description='List of process actions to process the input or output value.'
+    )
+
+    @validator('platform', always=True)
+    def create_empty_platform_list(cls, v):
         return [] if v is None else v
 
+    @validator('handler', always=True)
+    def check_duplicate_platform_name(cls, v, values):
+        v = [] if v is None else v
+        languages = [h.language for h in v]
+        dup_lang = find_dup_items(languages)
+        if dup_lang:
+            raise ValueError(
+                f'Duplicate use of language(s) found in alias handlers for '
+                f'{values["platform"]}: {dup_lang}. Each language can only be used once '
+                'in each platform.'
+            )
+        return v
 
-class DAGStringInput(DAGGenericInput):
-    """A String input.
+
+class DAGStringInputAlias(DAGGenericInputAlias):
+    """An Alias String input.
 
     You can add additional validation by defining a JSONSchema specification.
 
@@ -48,7 +98,7 @@ class DAGStringInput(DAGGenericInput):
 
     """
 
-    type: constr(regex='^DAGStringInput$') = 'DAGStringInput'
+    type: constr(regex='^DAGStringInputAlias$') = 'DAGStringInputAlias'
 
     default: str = Field(
         None,
@@ -67,8 +117,8 @@ class DAGStringInput(DAGGenericInput):
         return value
 
 
-class DAGIntegerInput(DAGGenericInput):
-    """An integer input.
+class DAGIntegerInputAlias(DAGGenericInputAlias):
+    """An alias integer input.
 
     You can add additional validation by defining a JSONSchema specification.
 
@@ -76,7 +126,7 @@ class DAGIntegerInput(DAGGenericInput):
     for more information.
 
     """
-    type: constr(regex='^DAGIntegerInput$') = 'DAGIntegerInput'
+    type: constr(regex='^DAGIntegerInputAlias$') = 'DAGIntegerInputAlias'
 
     default: int = Field(
         None,
@@ -95,15 +145,15 @@ class DAGIntegerInput(DAGGenericInput):
         return value
 
 
-class DAGNumberInput(DAGGenericInput):
-    """A number input.
+class DAGNumberInputAlias(DAGGenericInputAlias):
+    """An alias number input.
 
     You can add additional validation by defining a JSONSchema specification.
 
     See http://json-schema.org/understanding-json-schema/reference/numeric.html#numeric
     for more information.
     """
-    type: constr(regex='^DAGNumberInput$') = 'DAGNumberInput'
+    type: constr(regex='^DAGNumberInputAlias$') = 'DAGNumberInputAlias'
 
     default: float = Field(
         None,
@@ -122,7 +172,7 @@ class DAGNumberInput(DAGGenericInput):
         return value
 
 
-class DAGBooleanInput(DAGGenericInput):
+class DAGBooleanInputAlias(DAGGenericInputAlias):
     """The boolean type matches only two special values: True and False.
 
     Note that values that evaluate to true or false, such as 1 and 0, are not accepted.
@@ -132,7 +182,7 @@ class DAGBooleanInput(DAGGenericInput):
     See http://json-schema.org/understanding-json-schema/reference/boolean.html for more
     information.
     """
-    type: constr(regex='^DAGBooleanInput$') = 'DAGBooleanInput'
+    type: constr(regex='^DAGBooleanInputAlias$') = 'DAGBooleanInputAlias'
 
     default: bool = Field(
         None,
@@ -151,8 +201,8 @@ class DAGBooleanInput(DAGGenericInput):
         return value
 
 
-class DAGFolderInput(DAGGenericInput):
-    """A folder input.
+class DAGFolderInputAlias(DAGGenericInputAlias):
+    """An alias folder input.
 
     Folder is a special string input. Unlike other string inputs, a folder will be copied
     from its location to execution folder when a workflow is executed.
@@ -169,7 +219,7 @@ class DAGFolderInput(DAGGenericInput):
             "maxLength": 50,
         }
     """
-    type: constr(regex='^DAGFolderInput$') = 'DAGFolderInput'
+    type: constr(regex='^DAGFolderInputAlias$') = 'DAGFolderInputAlias'
 
     default: Union[HTTP, S3, ProjectFolder] = Field(
         None,
@@ -194,8 +244,8 @@ class DAGFolderInput(DAGGenericInput):
         return True
 
 
-class DAGFileInput(DAGFolderInput):
-    """A file input.
+class DAGFileInputAlias(DAGFolderInputAlias):
+    """An alias file input.
 
     File is a special string input. Unlike other string inputs, a file will be copied
     from its location to execution folder when a workflow is executed.
@@ -216,7 +266,7 @@ class DAGFileInput(DAGFolderInput):
         }
 
     """
-    type: constr(regex='^DAGFileInput$') = 'DAGFileInput'
+    type: constr(regex='^DAGFileInputAlias$') = 'DAGFileInputAlias'
 
     extensions: List[str] = Field(
         None,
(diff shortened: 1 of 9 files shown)


queenbee/io/alias.py
```diff
@@ -0,0 +1,568 @@
...
+    )
+
+    function: str = Field(
+        ...,
+        description='Name of the function. The input value will be passed to this '
+        'function as the first argument.'
+    )
+
+
+class DAGGenericInputAlias(GenericInput):
```
MingboPeng, line 45: Renaming it to DAGGenericBaseInputAlias?

queenbee/io/alias.py
```diff
@@ -0,0 +1,568 @@
...
+            "type": "string",
+            "maxLength": 50,
+            "pattern": "(?i)(^.*\\.epw$)"
+        }
+
+    """
+
+    type: constr(regex='^DAGStringInputAlias$') = 'DAGStringInputAlias'
+
+    default: str = Field(
```
MingboPeng, line 103: Why do we need the default here in alias object, can't we just use the default from upper non-alias object?

queenbee/io/alias.py
```diff
@@ -0,0 +1,568 @@
...
+    type: constr(regex='^DAGGenericInputAlias$') = 'DAGGenericInputAlias'
+
+    platform: List[str] = Field(
+        ...,
+        description='Name of the client platform (e.g. Grasshopper, Revit, etc). The '
+        'value can be any strings as long as it has been agreed between client-side '
+        'developer and author of the recipe.'
+    )
+
+    handler: List[IOAliasHandler] = Field(
```
MingboPeng, line 60: Renaming the key to handlers?

queenbee/io/alias.py
```diff
@@ -0,0 +1,568 @@
...
+
+class DAGGenericInputAlias(GenericInput):
+    """Base class for DAG Alias inputs.
+
+    This class adds a handler to input to handle the process of loading the input
+    from different graphical interfaces.
+    """
+    type: constr(regex='^DAGGenericInputAlias$') = 'DAGGenericInputAlias'
+
+    platform: List[str] = Field(
```
MingboPeng, line 53: Renaming the key to platforms?

queenbee/io/alias.py
```diff
@@ -0,0 +1,568 @@
...
+class DAGGenericOutputAlias(FromOutput):
+    """DAG generic alias output.
+
+    In most cases, you should not be using the generic output unless you need a dynamic
+    output that changes its type in different platforms because of returning different
+    objects in handler.
+    """
+    type: constr(regex='^DAGGenericOutputAlias$') = 'DAGGenericOutputAlias'
+
+    platform: List[str] = Field(
```
MingboPeng, line 430: Renaming the key to platforms?

queenbee/io/alias.py
```diff
@@ -0,0 +1,568 @@
...
+        ...,
+        description='Name of the client platform (e.g. Grasshopper, Revit, etc). The '
+        'value can be any strings as long as it has been agreed between client-side '
+        'developer and author of the recipe.'
+    )
+
+    handler: List[IOAliasHandler] = Field(
```
MingboPeng, line 437: Renaming the key to handlers?

queenbee/io/dag.py
```diff
@@ -343,12 +362,32 @@ def validate_spec(self, value):
...
 DAGInputs = Union[
-    DAGStringInput, DAGIntegerInput, DAGNumberInput, DAGBooleanInput,
+    DAGGenericInput, DAGStringInput, DAGIntegerInput, DAGNumberInput, DAGBooleanInput,
     DAGFolderInput, DAGFileInput, DAGPathInput, DAGArrayInput, DAGObjectInput
 ]
 
 
-class DAGFileOutput(FromOutput):
+class DAGGenericOutput(FromOutput):
+    """DAG generic output.
```
MingboPeng, line 371: """Base class for DAG output.

queenbee/io/dag.py
```diff
@@ -343,12 +362,32 @@ def validate_spec(self, value):
...
 
 DAGInputs = Union[
-    DAGStringInput, DAGIntegerInput, DAGNumberInput, DAGBooleanInput,
+    DAGGenericInput, DAGStringInput, DAGIntegerInput, DAGNumberInput, DAGBooleanInput,
     DAGFolderInput, DAGFileInput, DAGPathInput, DAGArrayInput, DAGObjectInput
 ]
 
 
-class DAGFileOutput(FromOutput):
+class DAGGenericOutput(FromOutput):
```
MingboPeng, line 370: Renaming it to DAGGenericBaseOutput?

tests/assets/recipes/baked/daylight-factor.yaml
```diff
@@ -34,25 +34,68 @@ flow:
...
+      name: model
+      description: null
+      default: null
+      spec: null
+      platform:
+      - rhino
+      handler:
+      - annotations: {}
+        type: IOAliasHandler
+        language: dotnet
```
MingboPeng, line 81: dotnet or csharp?

queenbee/io/alias.py
mostaphaRoudsari: This will be useful is the alias type is different from the original input. In those cases we need to be able to provide a default value from the same type of the alias input.

queenbee/io/alias.py
mostaphaRoudsari: I think we have kept most of the keys singular but let me check - I'm totally fine with changing this to handlers.

tests/assets/recipes/baked/daylight-factor.yaml
mostaphaRoudsari: This is a sample. I think csharp is the better choice. I will change this.

queenbee/io/common.py
```diff
@@ -14,12 +14,13 @@
 
 class ItemType(str, Enum):
     """Type enum for items in a list."""
+    Any = 'Any'  # a generic type to be used when different types can be used.
```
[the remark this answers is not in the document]
mostaphaRoudsari: What would you suggest in this case? This will be useful for dynamic inputs. In Grasshopper it should translate to No TypeHint

queenbee/io/dag.py
```diff
@@ -8,10 +8,29 @@
 
 from .common import ItemType, GenericInput, FromOutput
 from .artifact_source import HTTP, S3, ProjectFolder
+from .alias import DAGAliasInputs, DAGAliasOutputs
 from ..io.reference import FileReference, FolderReference, TaskReference
 
 
-class DAGStringInput(GenericInput):
+class DAGGenericInput(GenericInput):
```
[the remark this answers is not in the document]
mostaphaRoudsari: This can be used as an input type itself. Base will make it sound like a Baseclass that should be used directly.

queenbee/io/alias.py
mostaphaRoudsari: see 👆

queenbee/io/alias.py
mostaphaRoudsari: see 👆 - I will change this based on the decision that we make up there.

queenbee/io/alias.py
mostaphaRoudsari: see 👆 - I will change this based on the decision that we make up there.

AntoineDao (commented):
Standardising the handlers interface in terms of how it is intended to be "called" is a definite step in the right direction with regards to the previous PR. Thanks for that 😀

I am observing the following requirements from the alias/handler interface:
- Context: in what context should the handler be selected
- Operation: what handler operation/function should be used when translating the input into its Queenbee compatible type

### Flattening the Handlers
As such I am not sure how useful the `alias` key is except to discern between platforms. I think that we could add the `platform` key inside of each handler as an extra discriminant to make the structure flatter.

### RPC Abstraction Detail
I also appreciate the RPC (remote procedure call) like system where the handler "knows" what module and function to call for the transformation from platform input to queenbee input. I am concerned that the handler might have to know "too much" of the platform/GUI to make very specific calls. I understand this brings clear configurability, at the same time it feels like there is a risk of exposing too much such that it becomes difficult to change code that does not impact the handler API.

For example, imagine the code structure changes for the python `honeybee_rhino` where we move the `hb_model_to_hbjson` from the `handlers` folder into `handlers/daylight/`. From the user's perspective both the code and the API have changed, whereas from the developer's perspective this is just a code change. The result is still that the code below won't work anymore:

```yaml
handler:
- annotations: {}
  type: IOAliasHandler
  language: python
  module: honeybee_rhino.handlers
  function: hb_model_to_hbjson
```

Being overly specific might also remove some re-usability where a user could indicate "I want my grasshopper object to become a HBJSON file but I don't care if it's done with Python or C#".

### Change Proposal
**disclaimer**: I am not a fan of this proposal for the reasons explained above and I would rather we try something like I am proposing below. That being said if we can't reach an agreement then I'd say let's go for what @mostaphaRoudsari is proposing (maybe address the point about removing the `alias` key and just use a list of `handlers`). 

@mostaphaRoudsari , @MingboPeng , @chriswmackey and @cdriesler it would be helpful to have some input on your behalf regarding the proposed alternative below to manage object transformation into Queenbee inputs from different platforms (grasshopper, rhino, revit, web etc...)

I personally still feel that the handlers implementation is too specific for the reasons explained above. I would favour a system where we use the `annotations` on the `input` objects as an API to trigger the use of certain handlers. I would propose agreeing on a convention for the annotations such that the code running the handlers knows what to do.

Based on my first observation that what we need is `context` and `operation` I have a few proposals for how we might standardise input annotations. I have listed them below in order of my preference (you'll note it's from least specific to more specific 😅). You can compare the current system shown below to the examples provided for each standardisation proposal:

**Current Proposal**
```yaml
annotations: {}
type: DAGFileInput
name: model_hbjson
description: Honeybee model as a hbjson file.
alias:
- type: DAGGenericInputAlias
  name: model
  description: 'A path to a HBJSON file or a HB model object built with Python or dotnet libraries.'
  platform:
  - grasshopper
  handler:
  - type: IOAliasHandler
    language: python
    module: honeybee_rhino.handlers
    function: hb_model_to_hbjson
  - type: IOAliasHandler
    language: csharp
    module: HBRhino.Model
    function: HBModelToJSON
- type: DAGGenericInputAlias
  name: model
  platform:
  - rhino
  handler:
  - type: IOAliasHandler
    language: dotnet
    module: HBRhino.Model
    function: ExportRhinoModel
```
**Proposed Alternatives**
1. `<owner>: <operation>` : where the owner represents the overall org or group the object transformation process belongs to
```yaml
type: DAGFileInput
name: model_hbjson
description: Honeybee model as a hbjson file.
annotations:
  ladybug-tools: hb-model-to-json
```
2. `<owner>/<handler-name>: <operation>`: we add the handler name to further specify which library to call a function with. This can be helpful if functions names overlap (eg: "to-json")
```yaml
type: DAGFileInput
name: model_hbjson
description: Honeybee model as a hbjson file.
annotations:
  ladybug-tools/honeybee: hb-model-to-json
```
3. `<owner>/<platform>: <operation>`: we add the platform for some extra specificity which might help distinguish between owner operations if they overlap. Note that this could also be `<platform>/<owner>: <operation>`
```yaml
type: DAGFileInput
name: model_hbjson
description: Honeybee model as a hbjson file.
annotations:
  ladybug-tools/grasshopper: hb-model-to-json
  ladybug-tools/rhino: hb-model-to-json
  # alternative
  grasshopper/ladybug-tools: hb-model-to-json
  rhino/ladybug-tools: hb-model-to-json
```
4. `<owner>/<handler-name>.<platform>:  <operation>`: where the handler name represents a specific code library to use (ie: honeybee-rhino)
```yaml
type: DAGFileInput
name: model_hbjson
description: Honeybee model as a hbjson file.
annotations:
  ladybug-tools/honeybee.grasshopper: hb-model-to-json
  ladybug-tools/honeybee.rhino: hb-model-to-json
```
5. `<owner>.<handler-name>/<platform>.<language>: <operation>`: where we specific the language on top of the handler name in case of confusion.
```yaml
type: DAGFileInput
name: model_hbjson
description: Honeybee model as a hbjson file
annotations:
  "ladybug-tools.honeybee/grasshopper.c-sharp": "hb-model-to-json"
  "ladybug-tools.honeybee/grasshopper.python": "hb-model-to-json"
  "ladybug-tools.honeybee/rhino.c-sharp": "hb-model-to-json"
```

I think the main conceptual difference between this proposal and the work done by @mostapharoudsari is that using annotations enables users to express the behaviour they expect rather than how they want the transformation to happen under the hood. In this sense the `annotation` is used to `trigger` an action in the client library (Grasshopper plugin, Rhino Plugin, web interface, Revit plugin etc...) submitting the workflows.

This would then sheild the user from having to be too specific when declaring handlers and enable re-use of a specific annotation in mutliple contexts.

queenbee/io/common.py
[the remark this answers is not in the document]
MingboPeng: I think about this again, this "Any" should be fine as it is the value of an enum item, and it won't be changed to to "object".

queenbee/io/dag.py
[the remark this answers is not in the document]
MingboPeng: Ok, this is clear if it can be used directly for non-type specified input.

mostaphaRoudsari:
Thank you @AntoineDao for providing feedback.

## Flattening the handler

This can work but might end up in duplicate data. For instance with the current design you can use the same handler for `rhino` and `revit`.

## Proposed Alternatives

I like the simplicity of the proposal but there are two concerns:

1. It puts too much on the side of the UI developer and almost no control for the recipe developer. Basically recipe developer will always need the UI developer to support their proposal. I think the recipe author should be in control when it comes to handlers.
2. It doesn't provide a solution for the UI to know about the type, name or the description of the new input. I assume your proposal suggests that the UI developer should also make the decisions about these details?

I like to hear @MingboPeng's take to this proposal as he will be the first one to use it!

AntoineDao:
Ah yes you make a good point about duplication if we flatten the handlers! My bad and lack of foresight on that one 😅

### Proposed Alternative
I suspected that I didn't have the full picture in my head when making this proposal 😀

1. I think the configurability you get through the current proposal still requires a lot of work on the UI developer and little control on the recipe developer. By this I mean that the recipe developer is not expected to change the UI handler code and therefore exposing the code through the alias/handler interface doesn't offer much more control over using the annotation system proposed above.

2. I am not sure I fully understand the use case you describe here. My thinking is that the the plugin developper can determine the type of the input expected by Queenbee based on the input object directly and then use the annotation to determine which transformation to use. Some annotations might also be incompatible with certain input types. For example: "hb-model-to-json" would work for JSON, String and File inputs but not Bool, Folder or Int.

mostaphaRoudsari:
> I have also noticed an additional problem in both of our proposals 😀 How do we deal with the scenario where multiple competing handlers are set up? Ie: what happens if I put 3 Grasshopper/Python handlers in the input aliases/annotations?

We have a check for that one not to let it happen! :)

https://github.com/ladybug-tools/queenbee/pull/166/files#diff-289f89d906f4978609bad679fa670e40dfbac11d2ed825c98d5a51c39f7d06d9R69-R80

mostaphaRoudsari:
> That being said, I think that we might want to work on a plugin system for the UIs such that developers could write their own handlers which would be called based on the annotations in the file

I like to hear more about this. That's how I imagine the alias handler being used.

> My thinking is that the the plugin developper can determine the type of the input expected by Queenbee based on the input object directly and then use the annotation to determine which transformation to use. Some annotations might also be incompatible with certain input types. For example: "hb-model-to-json" would work for JSON, String and File inputs but not Bool, Folder or Int.

This is good thinking but how are thinking to communicate all these information with a annotation?

mostaphaRoudsari:
We decided to go with  the current alias system knowing that depending on how it pans out we might add support for annotations to make it easier for recipe developers to add handlers.

MingboPeng:
@mostaphaRoudsari since this is merged, can you add the openapi schema file to release assets?

mostaphaRoudsari:
Hi @MingboPeng, this is not merged to the master branch yet. The files that I gave you yesterday matches what we have here.

If we merge this in it will break the plugins. The sync component downloads the latest version from PYPI directly. We need to come up with a better strategy there so we don't end up with such limitations. cc: @chriswmackey

PR #170, opened 2020-11-11
Fix dependency model and add type to all Queenbee objects

### LINKED ISSUE #151: Generate "type" key for objects based on class name
When generating JSON/OpenAPI Schemas for dotnet the "type" key of certain objects should equal the model (class) name. We should probably automatically generate "type" keys from class names in the pydantic `BaseModel`  class.

### LINKED ISSUE #163: Conflict Types in Dependency
This DependencyType Type is going to conflict with the string Type that is used for tracking class name.

https://github.com/ladybug-tools/queenbee/blob/a059daa10ada9570ab4481e7c3a50099c4038d14/queenbee/recipe/dependency.py#L24-L27

cc @mostaphaRoudsari

mostaphaRoudsari:
This PR:

1. adds an extra check to ensure type field is always the same as the name of the object.
2. adds `type` field to all the Queenbee objects . depending on how to look at it resolves #151 
3. fix the objects that were using type for a different purpose. Resolves #163 
4. remove the from_folder test for baked recipes.


mostaphaRoudsari:
@MingboPeng, here is the asset files in case you want to test them on your side.
[queenbee.zip](https://github.com/ladybug-tools/queenbee/files/5525706/queenbee.zip)

tests/assets/recipes/baked/daylight-factor.yaml
```diff
@@ -1,5 +1,7 @@
+type: BakedRecipe
```
MingboPeng, line 1: Is this "BakedRecipe" a new class type or meant to be "Recipe"?

tests/assets/recipes/baked/daylight-factor.yaml
```diff
@@ -57,247 +62,257 @@ flow:
...
     description: null
     from:
-      annotations: {}
       type: TaskReference
+      annotations: {}
       name: post-process
       variable: post-process-folder
     alias: []
-  name: 8b959cb7cb3b5f24abd8ac64c03b44c457b318cc9494a7eeca514638edd43e34/main
+  name: 91d2c0727a0d749c710d71f5cbcee77cbc6a0cc2af29596f6ea236497eccf19b/main
```
MingboPeng, line 125: In Pollination's returned object, it is just "main" for the this entry flow, can we remove this digest for this main?

tests/assets/recipes/baked/daylight-factor.yaml
mostaphaRoudsari: BakeRecipe is different from Recipe. You can think about it as the compiled version of the recipe. You will still have a normal recipe that you can work with.

MingboPeng: It is correct in the regular recipe.

tests/assets/recipes/baked/daylight-factor.yaml
MingboPeng: it is main in the regular recipe.

mostaphaRoudsari: You should check the one under the valid folder. That doesn't have this.

queenbee/io/common.py
```diff
@@ -14,7 +14,7 @@
 
 class ItemType(str, Enum):
```
MingboPeng, line 15: Should we make enum items as same as the what we have for DAG inputs:
For example, here is missing: File, Path,
Renaming: Any =>Generic, Object => JSONObject

mostaphaRoudsari: That's a good idea. Will do that.

mostaphaRoudsari: https://github.com/ladybug-tools/queenbee/pull/170/commits/a1af031c1ef4bf6abf6e05eb70dcedde294389a8

MingboPeng: Did you forget "File", "Path"?

AntoineDao (commented):
Looking good overall. Setting the "type" for everything seems like a monumental effort but I guess there's not workaround with Pydantic for now... 😅

Only comment is about the `kind` key which I had understood would be set as a meaningful string in the serialized object and cast to a more precise name in the code version.

If it's a misunderstanding on my part and doing the reverse will not play nice with C# then let's have it this way for now 😀

queenbee/recipe/dependency.py
```diff
@@ -1,25 +1,27 @@
...
 
-class DependencyType(str, Enum):
+class DependencyKind(str, Enum):
+    """Dependency kind."""
+    recipe = 'recipe_kind'
 
-    recipe = 'recipe'
-
-    operator = 'operator'
+    operator = 'operator_kind'
```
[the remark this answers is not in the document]
mostaphaRoudsari: > Setting the "type" for everything seems like a monumental effort

Agreed! Luckily if the object is not used in a Union, the default value for type will be used. So technically we don't have to add type everywhere. Also that's how the tests are passing with minimal changes. I was surprised when they passed initially until I realized what is going on with the default value.

> ... the reverse will not play nice with C# then let's have it this way for now

So this one is C# specific as `operator` is a protected keyword in C# and, unfortunately, when translated to OpenAPI/dotnet it takes the keys and not the values. It also causing issues on the honeybee-schema. See: https://github.com/ladybug-tools/honeybee-schema/issues/210

When we rename operator to plugin we can get back to meaningful keywords.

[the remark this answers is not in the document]
AntoineDao: Shall we rename it to plugin now? This feels like a good occasion to do so 😀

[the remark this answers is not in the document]
mostaphaRoudsari: I'll do it in a separate PR. @MingboPeng, has asked for a number of other changes too. I'll do those and send a separate PR. I'm not going to miss updating Queenbee tests for a foreseeable future! 😀

PR #175, opened 2020-11-16
fix: change operator to plugin

mostaphaRoudsari:
This commit replaces the name `operator` with `plugin` which turned out to be a better term to communicate the concept. We have been using the term plugin instead of operator in the front-end for some time now.


MingboPeng (commented):
Thanks, looks all good to me.

AntoineDao (approved):
LGTM!

PR #181, opened 2020-11-21
fix(workflow): refactor workflow arguments

mostaphaRoudsari:
This commit refactors workflow arguments to match the changes in the rest of the library.

@MingboPeng, here is the updates schema files.

[queenbee-schema.zip](https://github.com/ladybug-tools/queenbee/files/5580464/queenbee-schema.zip)


queenbee/workflow/status.py
```diff
@@ -76,12 +76,12 @@ class TaskStatus(BaseStatus):
...
     )
 
-    inputs: Arguments = Field(
+    inputs: TaskArguments = Field(
         ...,
         description='The inputs used by this task'
     )
 
-    outputs: Arguments = Field(
+    outputs: TaskReturns = Field(
```
mostaphaRoudsari, line 84: I'm not sure if this change is the right change but since this is a task I feel the inputs and outputs should be TaskIO.

AntoineDao: I can't find a class called `TaskIO`. What should it look like and how would it differ from `TaskArguments` + `TaskReturns`?

AntoineDao (approved):
LGTM

PR #186, opened 2020-11-26
feat(status): add node status i/o primitives

AntoineDao:
These i/o types are used to link the original template input or output with the value generated by the task/workflow that has been run.

### DIFF
--- a/queenbee/io/inputs/node.py
+++ b/queenbee/io/inputs/node.py
@@ -1,0 +1,144 @@
+"""Queenbee input types for status nodes.
+
+For more information on plugins see plugin module.
+"""
+
+import os
+from typing import Union, List, Dict, Any
+from pydantic import constr, Field
+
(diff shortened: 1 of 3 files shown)


mostaphaRoudsari:
Not sure what a node is but that's a conversation for later. In any case make sure to check the code against flake8 before merging it in. I saw some cases here and there. 😀

AntoineDao:
Aaaah I accidentially comitted to `ladybug-tools` instead of `antoinedao`... Sorry about that. I did run `flak8` though and corrected some things here and there 😅

This [commit](https://github.com/ladybug-tools/queenbee/commit/6bdb7b3bc3c6e692d08acd2bfe09677221a4159e) should also be included in this PR. It fixes some IO translation bugs I caught on queenbee-argo and makes some names more helpful (renaming `tasks` to `nodes` on the `WorkflowStatus` object for example.

PR #204, opened 2020-12-08
feat(recipe): add RecipeInfo

mostaphaRoudsari:


queenbee/recipe/recipe.py
```diff
@@ -680,3 +680,38 @@ def template_by_name(templates: List[Union[DAG, TemplateFunction]], name: str) -
...
 
         return res
+
+
+class RecipeInfo(BaseModel):
+    """Recipe information.
+
+    Recipe information only includes metadata, inputs and outputs of a Recipe.
+    This object is useful for creating user interface for Recipes.
+    """
```
AntoineDao, line 690: Shouldn't there be a `type` key here?

AntoineDao (commented):
I agree that we need an object like this. I wonder however if we might slightly change the naming in order to satisfy both the current C# requirements and possible future recipe builder requirements.

### Functional Design
Exposing some the recipe's `inputs` and `outputs` serves the purpose of making it usable as a functional component. In this sense it is very similar to a plugin's `Function` object.

I wonder if we might benefit from changing `RecipeInfo` to `RecipeFunction` and equally create a corresponding `PluginFunction` which contains plugin metadata + a single function's `io` information.This will be the same as what you propose here and in addition we will have a clear path to search and use functions (from recipes or plugins) when building recipes using a visual builder.

AntoineDao:
After a call it seems we have agreed on the word `RecipeInterface` and this object should also contain a `source` key which is a url to where the recipe package can be downloaded.

queenbee/recipe/recipe.py
mostaphaRoudsari: Added!

mostaphaRoudsari:
@AntoineDao, I changed it to RecipeInterface and added the source key. Ready for another round of review.

queenbee/recipe/recipe.py
```diff
@@ -682,17 +682,24 @@ def template_by_name(templates: List[Union[DAG, TemplateFunction]], name: str) -
...
     This object is useful for creating user interface for Recipes.
     """
+    type: constr(regex='^RecipeInterface$') = 'RecipeInterface'
+
     metadata: MetaData = Field(
         ...,
         description='Recipe metadata information.'
     )
 
+    source: str = Field(
```
AntoineDao, line 698: Reckon we can use the [`AnyUrl`](https://pydantic-docs.helpmanual.io/usage/types/#urls) type from Pydantic here maybe?

AntoineDao (approved):
LGTM! Small detail in comments but we can merge as is.

queenbee/recipe/recipe.py
mostaphaRoudsari: I like using a more rigid type but AnyUrl doesn't allow `file:///` format.

PR #210, opened 2020-12-09
Status and version

### LINKED ISSUE #209: Add a new field for api_version
@AntoineDao, suggested that we should add a api_version to Queenbee similar to `apiVersion` tag in Argo: https://github.com/argoproj/argo/blob/master/examples/conditionals.yaml#L10

I'm not sure where do we want to add the field. It looks like that it should definitely be on a `recipe`. Maybe also on `functions` or should it be a single number for the `plugin` that will be inherited by all the `functions`?

AntoineDao:
This PR addresses the JobStatus I/O types issue and proposes an api versioning scheme set on specific object.

API versioning should only be set ofData Transfer Objects (DTOs) are any object that we expect to be used to be used by an external API (ie: not-queenbee) to be parsed and computed.

I have opted for a version and maturity flag for API versions (ie: version = v1 and maturity = beta1). This versioning scheme should ideally work like described for the Kubernetes project: https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups-and-versioning

fix #209

### DIFF
--- a/queenbee/job/job.py
+++ b/queenbee/job/job.py
@@ -10,6 +10,8 @@ class Job(BaseModel):
 
     A Job is an object to submit a list of arguments to execute a Queenbee recipe.
     """
+    api_version: constr(regex='^Job$') = 'Job'
+
     type: constr(regex='^Job$') = 'Job'
 
     source: str = Field(
--- a/queenbee/job/job.py
+++ b/queenbee/job/job.py
@@ -10,7 +10,7 @@ class Job(BaseModel):
 
     A Job is an object to submit a list of arguments to execute a Queenbee recipe.
     """
-    api_version: constr(regex='^Job$') = 'Job'
+    api_version: constr(regex='^v1beta1$') = 'v1beta1'
 
     type: constr(regex='^Job$') = 'Job'
 
--- a/queenbee/job/status.py
+++ b/queenbee/job/status.py
@@ -118,6 +118,8 @@ class StepStatus(BaseStatus):
 
 class JobStatus(BaseStatus):
     """Job Status."""
+    api_version: constr(regex='^v1beta1$') = 'v1beta1'
+
     type: constr(regex='^JobStatus$') = 'JobStatus'
 
     id: str = Field(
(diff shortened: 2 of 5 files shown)


queenbee/job/job.py
```diff
@@ -10,6 +10,8 @@ class Job(BaseModel):
 
     A Job is an object to submit a list of arguments to execute a Queenbee recipe.
     """
+    api_version: constr(regex='^Job$') = 'Job'
```
mostaphaRoudsari, line 13: This should probably be 'v1beta1'. What do you think about using an enumeration for api_versions?

```suggestion
    api_version: constr(regex='^v1beta1$') = 'v1beta1'
```

queenbee/job/status.py
```diff
@@ -132,7 +134,12 @@ class JobStatus(BaseStatus):
 
     steps: Dict[str, StepStatus] = {}
 
+    inputs: List[StepInputs] = Field(
```
mostaphaRoudsari, line 137: I thought the inputs should be DAGInputs  but if we can resolve the DAG inputs to StepInputs then this is just as good.

mostaphaRoudsari (commented):
Looks good! One small fix and a comment. Feel free to merge after fixing the api_version for Job.

queenbee/job/job.py
AntoineDao: Thanks for the fix! So my understanding was that Enumerations are not allowed because of C#? I would consider this item to be the same as the `type` key. It's a fixed value for this object and helps discern it from other objects.

queenbee/job/status.py
AntoineDao: DAG Inputs don't contain any of the actual data used. A DAG input only has a schema and a default value. So in this case I was hesitating between `JobArguments` and `StepInputs` and decided to go for the `StepInputs` because they contain both the actual value and the schema information.

mostaphaRoudsari: You are correct. I meant JobArguments and I agree with your decision.

queenbee/job/job.py
mostaphaRoudsari: Enums are fine in C#. It was just we couldn't use some specific keys.

I'm thinking that in long-term we want to be able to support more than one. We can add the enum at that point. If we do now it will only have a single value.

mostaphaRoudsari (approved):
LGTM! Thanks.

PR #211, opened 2020-12-09
fix(license): remove license file

### LINKED ISSUE #154: Change license to be a link instead of a file
Currently the idea is that each operator or recipe we will include a license file - similar to what GitHub provides. I propose that we change how Queenbee license is assigned to operators and recipes and instead of copying the file, provide a link/url to it. Here are the two main reasons:

1. In 99% of the cases the licenses are not edited and it is really redundant to copy them over with each recipe or operator.

2. The link will also make it easy to include the licensing information in the executable versions of the code. For instance, queenbee-luigi can add a header that the file is generated based on XYZ and is licensed under ABC with a link to the license. This is specially helpful if the access to the original recipe is not available for some reason and one cannot see the original license file.

mostaphaRoudsari:
This PR removes the license file and uses the license field in meta data instead as the single source of declaring a license for a plugin or a recipe.

Resolves #154

### DIFF
--- a/queenbee/plugin/plugin.py
+++ b/queenbee/plugin/plugin.py
@@ -1,4 +1,5 @@
 """Queenbee Plugin class."""
+import json
 import os
 import yaml
 from typing import List
@@ -143,7 +144,7 @@ class Plugin(BaseModel):
 
         return cls.parse_obj(plugin)
 
-    def to_folder(self, folder_path: str, readme_string: str = None, license_string: str = None):
+    def to_folder(self, folder_path: str, *, readme_string: str = None):
         """Write a plugin to a folder
 
         Note:
@@ -164,7 +165,6 @@ class Plugin(BaseModel):
 
         Keyword Arguments:
             readme_string {str} -- The README file string (default: {None})
-            license_string {str} -- The LICENSE file string (default: {None})
         """
         os.makedirs(os.path.join(folder_path, 'functions'), exist_ok=True)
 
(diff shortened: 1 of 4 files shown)


queenbee/plugin/plugin.py
```diff
@@ -186,6 +186,5 @@ def to_folder(self, folder_path: str, readme_string: str = None, license_string:
             with open(os.path.join(folder_path, 'README.md'), 'w') as f:
                 f.write(readme_string)
 
-        if license_string is not None:
-            with open(os.path.join(folder_path, 'LICENSE'), 'w') as f:
-                f.write(license_string)
+        with open(os.path.join(folder_path, 'LICENSE'), 'w') as f:
```
AntoineDao, line 189: Should we just drop the LICENSE file if we're not actually going to write anything there?

AntoineDao (commented):
~~Yep the change makes sense. Reckon we can probably just drop anything to do with a file called `LICENSE` 😀~~

AntoineDao (approved):
Yep the change makes sense. Reckon we can probably just drop anything to do with a file called `LICENSE` 😀 If you want to keep the file for some reason the you can still consider this PR approved!

queenbee/plugin/plugin.py
mostaphaRoudsari: I think it might still be helpful for user discovery. We can remove it at any point and it shouldn't break anything since it is only generated insider the folder when one uses to_folder.

PR #250, opened 2021-01-23
add optional artifact

mostaphaRoudsari:
This commit adds support for optional artifact inputs and outputs. This change will make it possible to support radiance recipes with BSDF materials.

@AntoineDao, I added an `is_optional` property to artifact input/outputs which can be used to update the extensions to support optional types.


AntoineDao (approved):
I'll merge this in just now so I can implement it on queenbee-argo ASAP.

PR #264, opened 2021-02-11
docs(guides): update guides

tymokvo:
# What
- Minor style/clarity updates to `guides/plugin`


docs/guides/recipe.rst
```diff
@@ -10,37 +10,102 @@ In this section we will walk you through how to create a new Recipe.
...
+the ``dependencies`` array should point to a valid, packaged Plugin at a local
+path (such as the result from the ``Create a New Recipe`` documentation) or
+hosted on a publicly-available package server.
 
 Flow
 ----
 
+The ``flow/`` directory contains the YAML definitions of the DAGs of the recipe
+as YAML files. At a minimum, the directory must contain a ``main.yaml`` which
+is, surprisingly, the main graph.
```
AntoineDao, line 110: 🤯

docs/guides/recipe.rst
```diff
@@ -94,6 +115,178 @@ DAG Tasks
 As mentioned above, a Task can itself be a DAG. This both allows comlpex graphs
 to be reused inside other graphs and allows isolating related Tasks into a logical unit.
 
+Add a new file at the path ``flow/dag.yaml``. Inside that file, place the following snippet:
```
AntoineDao, line 118: Did you mean `flow/main.yaml`?

AntoineDao:
Had a quick look despite the `draft` status and thank you! 😍 Very happy to have another chat about some concepts you feel are shaky or unclear after you've had a chance to take a first look at queenbee.

docs/guides/recipe.rst
tymokvo: I was thinking that the documentation example should make use of both, a `dag.yaml` that is referenced in the `main.yaml` to show how a nested DAG can be consumed in a Recipe.

tymokvo:
Thanks for checking it out! I figured I'd take it as far as I could from reading the source and examples in `private-garden` until getting stuck. I should be able to finish it after talking to Mostapha today.

docs/guides/recipe.rst
AntoineDao: Ah right I get you. I would call it `sub-dag` or something that gives the idea that it is not the main entrypoint of the recipe. Thoughts @mostaphaRoudsari ?

mostaphaRoudsari: I'm going to put an example here to iterate over and energy simulation and we can call it based on the name of it's functionality. I'll do a first pass and you can keep it if you like it.

AntoineDao:
By the way @tymokvo feel free to `un-draft` this PR and merge it in. It would be good to have functioning docs again sooner rather than later 😅

mostaphaRoudsari:
Agreed! This is already much better than what we have. We can have a follow up PR with additional items.

tymokvo:
Ok! I will merge it in its current form, and then could someone open an issue to improve the areas that I couldn't fill in? I will happily close the issue when I know how to do so. :smile_cat:

tymokvo:
Oh, I don't have write access so someone else will have to merge it :crying_cat_face:

mostaphaRoudsari:
@tymokvo try again now! Changed the access for backend team.

PR #268, opened 2021-02-18
fix: resolve local dependency path on \*nix correctly

### LINKED ISSUE #266: Local dependency path resolution can fail on *nix
Steps to reproduce

1. Create new virtual env and activate
2. Install a local `pollination-dsl` dependency
3. Run `pollination dsl translate <name of recipe using local dependency> <some local path>`

Expect:
Command fails with:
```console
urllib.error.URLError: <urlopen error [Errno 2] No such file or directory: '//<current working directory>/<absolute path to index.json>'>
```

This appears to come from [this line](https://github.com/pollination/queenbee/blob/4f9d3588753950b401fd6e97a4f26811eb986d46/queenbee/recipe/dependency.py#L89) where the assumption is made that, if the path starts with `file://`, then it is relative to the current working directory.

mostaphaRoudsari:
This PR addresses the issue with resolving the local path as a uri. It took mixed usage of `os.path` and `pathlib.Path` to get it right on both Windows and Linux. Part of the issue was how in Windows and \*nix systems based on what the uri is pointing to the number of slashes after file:// can be 2 or 3! I was expecting it to always be 3 which would return `home/user/...` instead of `/home/user/...`. The first one is considered an absolute path and the second one is relative! That's the main reason that it was failing. I also took the opportunity to clean the code a little bit.

resolves #266

### DIFF
--- a/queenbee/base/request.py
+++ b/queenbee/base/request.py
@@ -1,25 +1,40 @@
-import pathlib
 import os
+import pathlib
+import platform
 from urllib import request
-from typing import Union, Dict
-from .basemodel import BaseModel
+from typing import Dict
 
 USER_AGENT_STRING = 'Queenbee'
 
 
 def get_uri(url):
     """Resolve uri for urls and local files."""
-
-    if url.startswith('file:///'):
+    if url.startswith('file://'):
         return url.replace('\\', '/')
     elif url.startswith('http://') or url.startswith('https://'):
         return url
 
-    # a local file 
+    # a local file but is not formatted as local url
+    pre = 'file:///' if platform.system() == 'Windows' else 'file://'
+    return resolve_local_source(pre + url)
+
+
+def resolve_local_source(path, as_uri=True):
+    """Get an absolute path for a file:// or file:/// path.
+
+    Apparently both of them are used: https://en.wikipedia.org/wiki/File_URI_scheme
+    """
+    sep = 'file:///' if platform.system() == 'Windows' else 'file://'
+
     try:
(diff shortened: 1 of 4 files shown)


queenbee/base/request.py
```diff
@@ -1,25 +1,40 @@
...
-
-    if url.startswith('file:///'):
+    if url.startswith('file://'):
         return url.replace('\\', '/')
     elif url.startswith('http://') or url.startswith('https://'):
         return url
 
-    # a local file 
+    # a local file but is not formatted as local url
+    pre = 'file:///' if platform.system() == 'Windows' else 'file://'
```
tymokvo, line 18: Since this logic is duplicated a couple times and cross-platform compatibility is so important here, what would you think about moving this to a small `env` module that handles setting the OS indicator once across the whole package?

tymokvo, line 18: Since this logic is duplicated a couple times and cross-platform compatibility is so important here, what would you think about moving this to a small `env` module that handles setting the OS indicator once across the whole package?

tymokvo (approved):
This worked great, thank you @mostaphaRoudsari ! I left one comment about abstracting some OS logic but I don't think it's key to shipping this and I would also be happy to take it in a follow up PR, if you'd like.

queenbee/base/request.py
mostaphaRoudsari: Good idea! I created an issue here so we don't forget about it: https://github.com/pollination/queenbee/issues/269 😀

PR #270, opened 2021-02-19
feat(env): Add env module

### LINKED ISSUE #269: Create an env module to handle setting the OS indicator
@tymokvo: Since this logic is duplicated a couple times and cross-platform compatibility is so important here, what would you think about moving this to a small env module that handles setting the OS indicator once across the whole package?

See here: https://github.com/pollination/queenbee/pull/268#discussion_r578696698

tymokvo:
# What
- Add `env` module for environment-specific values
- Use in `request` module for local files

# Why
- To DRY up the code
- To fix #269

### DIFF
--- a/queenbee/env/os.py
+++ b/queenbee/env/os.py
@@ -7,4 +7,4 @@ class OS:
     is_windows: bool = platform.system() == 'Windows'
     is_nix: bool = platform.system() == 'Linux'
 
-    file_uri_prefix = 'file:///' if is_windows else 'file://'
+    file_uri_prefix: str = 'file:///' if is_windows else 'file://'
--- a/queenbee/env/os.py
+++ b/queenbee/env/os.py
@@ -1,0 +1,10 @@
+import platform
+
+
+class OS:
+    """A simple wrapper to handle OS-specific switches."""
+
+    is_windows: bool = platform.system() == 'Windows'
+    is_nix: bool = platform.system() == 'Linux'
+
+    file_uri_prefix = 'file:///' if is_windows else 'file://'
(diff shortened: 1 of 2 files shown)


queenbee/env/os.py
```diff
@@ -0,0 +1,18 @@
...
+
+    Attributes:
+        is_windows: A boolean of whether the current system is MS Windows.
+        is_nix: A boolean of whether the current system is *nix (any Linux).
+        file_uri_prefix: A string to be used as the default prefix for URIs
+            which point to local files on the system.
+
+    """
+
+    is_windows: bool = platform.system() == 'Windows'
```
mostaphaRoudsari, line 15: This attribute not being read-only can be tricky but since this is a simple wrapper I think we are good in this case.

queenbee/env/os.py
```diff
@@ -0,0 +1,18 @@
...
+    is_nix: bool = platform.system() == 'Linux'
```
mostaphaRoudsari, line 16: What if we set this one to `not is_windows` to also catch the case for running on Mac? I think it returns `Darwin` on Mac.

```suggestion
    is_nix: bool = not is_windows
```

mostaphaRoudsari (commented):
This is much cleaner! Thanks. Just a small comment about catching the case on Mac computers. It's an easy one since we really only care if it is Windows or not. Feel free to merge this in after addressing that comment.

Also I would consider this a `fix` more than a `feat` but your approach is technically correct! We are adding a new module. I'm just sharing my wrong temptations in public in order to be better in the future. :)

tymokvo:
Thanks for the feedback @mostaphaRoudsari ! Requested again in case there's anything else you think should be changed.

mostaphaRoudsari (approved):
LGTM! Thank you @tymokvo!

PR #271, opened 2021-02-20
fix(guides): update recipe example

### LINKED ISSUE #198: Update docs to match the new Queenbee schema
The docs are written for the previous iteration of the schema - they need to be updated.

tymokvo:
# What
- Update recipe example to be helpful

# Why
- To fix #198 
- To be good OSS maintainers


tymokvo:
@mostaphaRoudsari I started updating this with the example that you prepared, but I am worried that it is too verbose and requires too many outside steps (like having access to the the energy-plus plugin that you defined with `pollination-dsl`). The yaml snippets alone are huge and I think would be off-putting to me if I were coming to this as an outsider.

What do you think about the example using something simpler like the `walk in -> take off shoes -> ...` sample? I wanted the example to be more practical but I think the recipe will just be too large.

mostaphaRoudsari:
Haha! That's what I was worried about. If you can come up with a simple meaningful example I'll be happy to help. Maybe we can get some ideas by looking what Argo is using for their examples?

That being said, I'm not sure how many people will write a queenbee recipe from scratch in yaml format. They will most likely use the DSL or the visual interface to build those recipes. If we think about it like that then having a real world example that they can copy paste and run with some preparation might actually be a better example than an abstract example that doesn't deal with the real world problem.

I can go either way. You let me know which one do you prefer!

docs/guides/recipe.rst
```diff
@@ -110,9 +110,170 @@ the Recipe.
...
+      required: true
+      spec: null
+    outputs: []
+    name: f634208a54c93a7f08a38336187e7e17b437bd929e4f56d3b16b7b9510b496cb/main
+    fail_fast: true
+    tasks:
+    - type: DAGTask
+      annotations: {}
+      name: energy-simulation-parallel-run
+      template: f634208a54c93a7f08a38336187e7e17b437bd929e4f56d3b16b7b9510b496cb/ep-run
```
mostaphaRoudsari, line 229: Hi @tymokvo, I know this is a draft PR but I thought I should point this out and save you some time. This YAML is from a baked-recipe. In a non-baked `Recipe` the template is referred to by name instead of hash and Queenbee resolves those dependencies when "baking" them. That keeps the original recipes human-readable and fixes the dependencies in the baked version. Here is the recipe as a folder - not being baked.

[parametric-energy-run.zip](https://github.com/pollination/queenbee/files/6015260/parametric-energy-run.zip)

You can also generate it yourself if you want. I didn't think of a case where someone wants to generate the non-baked version of the recipe but it is possible to do. You can use `--dry-run` option of `push` command to get a non-baked version of the recipe. It prints out the path to where the recipe has been exported to.

`pollination dsl push pollination-parametric-energy-run --dry-run`

tymokvo: Ah, thank you!

tymokvo:
So I think my inclination is to just use the real recipe that can be copied and pasted with a disclaimer in the preamble that explains that making new recipes should be done with the DSL. It will still be long, but at least it will be fully explained for a determined reader.

It seems like that should be copy-and-pastable with the ZIP you attached above since the plugin is available in the pollination repo, correct?

mostaphaRoudsari:
> It seems like that should be copy-and-pastable with the ZIP you attached above since the plugin is available in the pollination repo, correct?

The plugin is not actually available in the repo. We can make it available but I wasn't sure if @chriswmackey will be happy with me using pollination-energyplus namespace for the sample recipe. If we can put a sample or something in the name of the recipe we can create two new repos for them and even push them to Pollination. I _feel_ if we can find a way to make them docs only that might be the safer approach for now.


And I agreed that a practical complex example is better for someone who wants to learn Queenbee than something that is too abstract for people to make the connection.

tymokvo:
Ok, ready for a real review this time @mostaphaRoudsari !

docs/guides/recipe.rst
```diff
@@ -69,21 +74,52 @@ non-executable fields of ``package.json`` such as ``name``, ``version``,
 * **keywords**: A list of strings that are related to the package.
 * **maintainers**: A list of ``Maintainer`` objects that describe the contact
```
mostaphaRoudsari, line 75: We might want to link the objects like Maintainer to its schema so one can see how to create one: https://pollination.github.io/queenbee/_static/redoc-recipe.html#tag/maintainer_model

docs/guides/recipe.rst
```diff
@@ -69,21 +74,52 @@ non-executable fields of ``package.json`` such as ``name``, ``version``,
...
   information of the people responsible for the recipe (you, in this case).
-* **home**: ?
-* **sources**: ?
-* **icon**: A publicly-accessible URI to give the recipe a friendly face.
+* **home**: The homepage for this Recipe.
+* **sources**: A list of necessary container image URIs.
```
mostaphaRoudsari, line 78: This field is similar to Python's extra links - it can be the link to the container image but it can also be link to documentation or source code and so on.

docs/guides/recipe.rst
```diff
@@ -69,21 +74,52 @@ non-executable fields of ``package.json`` such as ``name``, ``version``,
...
+* **icon**: A publicly-accessible image URI to give the recipe a friendly face.
 * **deprecated**: Whether or not the recipe is in active development.
 * **description**: A textual description, analogous to a README.
 * **license**: A valid `SPDX Identifier <https://spdx.org/licenses/>`_ which
```
mostaphaRoudsari, line 82: This has changed: https://pollination.github.io/queenbee/_static/redoc-recipe.html#tag/license_model

docs/guides/recipe.rst
```diff
@@ -98,46 +134,183 @@ with the following snippet:
...
+        variable: sensor-count
+    - type: TaskArgument
+      annotations: {}
+      name: radiance-parameters
+      from:
+        type: InputReference
+        annotations: {}
+        variable: radiance-parameters
+    - type: TaskPathArgument
+      annotations: {}
```
mostaphaRoudsari, line 210: For the sake of readability we may want to remove some of these unset fields like annotations.

docs/guides/recipe.rst
```diff
@@ -220,9 +394,96 @@ to the first element of ``create-rad-folder``'s ``returns`` array.
...
+engine, locally with Luigi, or in the cloud with Pollination, the execution
+engine is free to locate the outputs from this task as it sees fit. The
+referenced paths are simply relative to the execution context where the task is
+actually executed which allows the same Recipe to be used locally for
+convenience or in the cloud for enabling massive scale without changes.
+
+Luigi, for instance, will create a temporary folder on your local drive unique
+to each task which will become the context for resolving the path while
+Pollination will run the task inside a container and that container's
+filesystem will become the context for path resolution.
```
mostaphaRoudsari, line 410: YES! That's how I should have explained it to people. 😄

mostaphaRoudsari (commented):
Nicely done @tymokvo! I added one suggestion and one note about the LICENSE object that has changed.

From reading the notes I also learned about how I can explain the execution logic to other people in the future. Thank you for that! :)

mostaphaRoudsari (approved):
LGTM! Thank you @tymokvo!

You win the OSS maintainer award of the month because of this PR!

tymokvo:
> You win the OSS maintainer award of the month because of this PR!

A highly coveted award! I'm honored :smile_cat:

Issue #280, opened 2021-02-27
Pydantic 1.8 breaks __hash__ functions


AntoineDao:
I noticed a bug on `honeycomb-simulations` whereby some class instances aren't hashing properly. It seems to be linked to the latest version of queenbee where pydantic was upgraded to 1.8. I noticed people reporting [similar bugs](https://github.com/samuelcolvin/pydantic/issues/2422) already.

I will fix honeycomb-simulations manually just now and open a PR to queenbee to revert back to pydantic==1.7.3.

PR #281, opened 2021-02-27
fix(deps): revert to pydantic 1.7.3

AntoineDao:
Pydantic 1.8 breaks the hashing functions somehow. I think we should move to 1.8 once they have
fixed that bug. I am also not sure how to go about testing this bug reliably to avoid future
dependencies update that cause a similar bug...

#280

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-pydantic==1.8
+pydantic==1.7.3
 PyYAML==5.4.1
 jsonschema==3.2.0
--- a/tests/base/hash_test.py
+++ b/tests/base/hash_test.py
@@ -1,0 +1,18 @@
+import pytest
+
+from ._base import BaseTestClass
+
+@pytest.mark.hash
+class BaseHashTest(BaseTestClass):
+
+    def pytest_generate_tests(self, metafunc):
+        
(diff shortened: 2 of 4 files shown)


mostaphaRoudsari:
Hi @AntoineDao, I think a first good step is to remove auto-merge for queenbee so we can do a double check before merging stuff in.

https://github.com/pollination/queenbee/blob/fa0d18a373c79658d66b818aa4c29b4b5055a656/.dependabot/config.yml#L6-L9

For this specific issue we can add a test to check the hash for a simple recipe which is not going to be updated - which I assume means will always have the same hash. I'm not sure how to add a normal single test to the fancy parametrized test that you have set up here otherwise I would tried to add one.

mostaphaRoudsari:
Also it looks like there will be a 1.8.1 release soon: https://github.com/samuelcolvin/pydantic/pull/2423#issuecomment-787092042

AntoineDao:
I am going to merge this in so I can fix the backend dependencies as needed. I have added basic tests to ensure that the __hash__ function does not return `None` as was the case for pydantic 1.8. I think this is better than changing the `automerge` on dependabot as if the tests fail dependabot will not merge 😀

PR #298, opened 2021-03-10
fix(status): enumerate job/run/step status strings

AntoineDao:
This makes it clearer what status strings to expect and helps generate more robust SDK clients.

### DIFF
--- a/queenbee/job/job.py
+++ b/queenbee/job/job.py
@@ -55,6 +55,8 @@ class JobStatusEnum(str, Enum):
     running = 'Running'
     # The job has failed to schedule runs
     failed = 'Failed'
+    # The job has been cancelled by a user
+    cancelled = 'Cancelled'
     # All runs have either succeeded or failed
     completed = 'Completed'
     # Could not determine the status of the job
--- a/queenbee/job/job.py
+++ b/queenbee/job/job.py
@@ -1,3 +1,4 @@
+from enum import Enum
 from typing import Dict, List, Generator
 from datetime import datetime
 
@@ -5,7 +6,6 @@ from pydantic import Field, constr
 
 from ..base.basemodel import BaseModel
 from ..io.inputs.job import JobArguments
-from .status import BaseStatus
 
 class Job(BaseModel):
     """Queenbee Job.
@@ -44,6 +44,23 @@ class Job(BaseModel):
     )
 
 
+class JobStatusEnum(str, Enum):
+    """Enumaration of allowable status strings"""
+
+    # The job has been created
+    created = 'Created'
+    # The job folder is being prepared for execution and runs are being scheduled
+    pre_processing = 'Pre-Processing'
+    # Runs have been scheduled
+    running = 'Running'
+    # The job has failed to schedule runs
+    failed = 'Failed'
+    # All runs have either succeeded or failed
+    completed = 'Completed'
+    # Could not determine the status of the job
+    unknown = 'Unknown'
+
+
 class JobStatus(BaseModel):
     """Parametric Job Status."""
 
@@ -56,20 +73,19 @@ class JobStatus(BaseModel):
         description='The ID of the individual job.'
     )
 
-    status: str = Field(
-        ...,
-        description='The status of this task. Can be "Running", "Succeeded", "Failed" '
-        'or "Error"'
+    status: JobStatusEnum = Field(
+        JobStatusEnum.unknown,
+        description='The status of this job.'
     )
 
     message: str = Field(
         None,
-        description='Any message produced by the task. Usually error/debugging hints.'
+        description='Any message produced by the job. Usually error/debugging hints.'
     )
 
     started_at: datetime = Field(
         ...,
-        description='The time at which the task was started'
+        description='The time at which the job was started'
     )
 
     finished_at: datetime = Field(
--- a/queenbee/job/run.py
+++ b/queenbee/job/run.py
@@ -14,6 +14,23 @@ from ..io.outputs.step import StepOutputs
 from .status import BaseStatus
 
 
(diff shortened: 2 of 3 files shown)


queenbee/job/job.py
```diff
@@ -44,6 +44,23 @@ class Job(BaseModel):
     )
 
 
+class JobStatusEnum(str, Enum):
+    """Enumaration of allowable status strings"""
+
+    # The job has been created
+    created = 'Created'
+    # The job folder is being prepared for execution and runs are being scheduled
+    pre_processing = 'Pre-Processing'
```
tymokvo, line 53: What do you think about keeping them all lowercase and punctuation free? It seems easy to forget what the convention is in the case of doing a direct string comparison. (E.g is it `preprocessing` or `Pre-Processing`?) But I can be convinced otherwise!

tymokvo (commented):
Overal this is looking good to me! I added one small comment enum value convention.

queenbee/job/job.py
AntoineDao: Am I right in thinking you perceive this as an issue for developers using the Python Queenbee library? 

My thinking is that once you start using `Enum` you should be using the Enum class for string comparison instead of direct string comparison. This means that the enum "raw" value is different from the "conceptual" value it represents in your code. This should in theory make things more robust.

```python
status_string = 'Pre-Processing'

# Brittle
# Direct string comparison
if status_string == 'Pre-Processing':
  do_stuff()

# More robust
# Enum value based comparison
if status_string == JobStatusEnum.pre_processing:
  do_stuff()
```

mostaphaRoudsari: I agree with the comparison but there is another reason that I think removing the `-` is a good idea. When enums are translated to C# through (Pydantic -> OpenAPI -> Code generator -> Queenbee .NET) it only takes the value - not the name during the translation. Foe the value to be value we should remove the `-` so either postprocessing or PostProcessing are good options. @MingboPeng, let me know if this has changed from the last time that we talked about.

Also what happened to `canceled` status?

tymokvo: Yep, I was thinking of the python library only. And I agree that one should be using the more robust enum value comparison. My point was that, in either case, I (a hypothetical consumer) need to remember what the convention is for the string literal that is put into `status_string`.

@mostaphaRoudsari brings up a good point about the code generation.

AntoineDao:
Urgh I forgot about the `Cancelled` status when deleting the `Paused` one 🙃 My bad @mostaphaRoudsari sorry.

queenbee/job/job.py
AntoineDao: I am going to wait for @MingboPeng to comment on this before merging but I just ran the OpenAPI generator for the Queenbee C# SDK repo and `Enums` clean up the string for the value name and render like this:

```c#
/* 
 * Queenbee Schema with Inheritance
 *
 * Documentation for Queenbee schema.
 *
 * Contact: info@ladybug.tools
 * Generated by: https://github.com/openapitools/openapi-generator.git
 */

using System;
using System.Linq;
using System.IO;
using System.Text;
using System.Text.RegularExpressions;
using System.Collections;
using System.Collections.Generic;
using System.Collections.ObjectModel;
using System.Runtime.Serialization;
using Newtonsoft.Json;
using Newtonsoft.Json.Converters;
using System.ComponentModel.DataAnnotations;


namespace QueenbeeSDK
{
    /// <summary>
    /// Enumaration of allowable status strings
    /// </summary>
    /// <value>Enumaration of allowable status strings</value>
    
    [JsonConverter(typeof(StringEnumConverter))]
    
    public enum RunStatusEnum
    {
        /// <summary>
        /// Enum Created for value: Created
        /// </summary>
        [EnumMember(Value = "Created")]
        Created = 1,

        /// <summary>
        /// Enum Scheduled for value: Scheduled
        /// </summary>
        [EnumMember(Value = "Scheduled")]
        Scheduled = 2,

        /// <summary>
        /// Enum Running for value: Running
        /// </summary>
        [EnumMember(Value = "Running")]
        Running = 3,

        /// <summary>
        /// Enum PostProcessing for value: Post-Processing
        /// </summary>
        [EnumMember(Value = "Post-Processing")]
        PostProcessing = 4,

        /// <summary>
        /// Enum Failed for value: Failed
        /// </summary>
        [EnumMember(Value = "Failed")]
        Failed = 5,

        /// <summary>
        /// Enum Succeeded for value: Succeeded
        /// </summary>
        [EnumMember(Value = "Succeeded")]
        Succeeded = 6,

        /// <summary>
        /// Enum Unknown for value: Unknown
        /// </summary>
        [EnumMember(Value = "Unknown")]
        Unknown = 7

    }

}
```

So it still seems ok for us to use `Post-Processing` as a value. It's worth noting I have an interest in keeping it that way for 2 reasons:
- An enum value is a value that shouldn't be concerned with how code handles it (ie: we should be allowed to have `illegal` string types in an enum)
- The database is currently filled with strings in the format above and I'd rather not run another migration 🙃

MingboPeng: Yes, OpenAPI generator will remove all special characters during the translation. But I would vote for the "PostProcessing" so that it keeps consistent across different languages.
CSharp:
```csharp
JobStatusEnum.PostProcessing 
```

Python:
```python
JobStatusEnum.PostProcessing 
```

MingboPeng (commented):
Thanks @AntoineDao, it looks good to me.

AntoineDao:
Thanks for the feedback folks. I am going to go ahead and keep it as "Post-Processing" instead of converting it to PascalCase because casing is language specific (ie: Python uses `snake_case`, C# uses `PascalCase` and Javascript uses `camelCase` so there will never be a "consistent" enum key name 😅).

PR #318, opened 2021-03-30
feat(packages): Use AnyUrl type

tymokvo:
I noticed while making search filters for packages that there are several fields that are described as URLs but don't use the `pydantic.AnyUrl` type. Though, maybe this was intentional.

### DIFF
--- a/queenbee/base/metadata.py
+++ b/queenbee/base/metadata.py
@@ -71,17 +71,17 @@ class MetaData(BaseModel):
         description='A list of maintainers for the package'
     )
 
-    home: str = Field(
+    home: AnyUrl = Field(
         None,
         description='The URL of this package\'s home page'
     )
 
-    sources: List[str] = Field(
+    sources: List[AnyUrl] = Field(
         None,
         description='A list of URLs to source code for this project'
     )
 
-    icon: str = Field(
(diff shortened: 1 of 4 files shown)


mostaphaRoudsari:
Hi @tymokvo, thank you for the PR. Good observation. 

For metadata I wanted to keep the fields flexible for users. We can change that if you think that's better.

For `recipes` and `dependencies` it was intentional. Unfortunately `AnyUrl` doesn't support a `uri` for local files. That means we couldn't use local registries with a filepath. Here is a simple example for you to test.

```python
class Package(BaseModel):
    source: AnyUrl

Package(source='file:///C:/Users/username/.queenbee')
```

```
pydantic.error_wrappers.ValidationError: 1 validation error for Package
source
  URL host invalid (type=value_error.url.host)
```

tymokvo:
Got it! That is a slightly tricky caveat. The `*nix` equivalent of your example does work:

```python
class A(BaseModel):
    u: AnyUrl

print(A(u='file://localhost/home/tyler/.bashrc'))

>>> A(u=AnyUrl('file://localhost/home/tyler/.bashrc', scheme='file', host='localhost', host_type='int_domain', path='/home/tyler/.bashrc'))
```

PR #343, opened 2021-05-26
feat(handler): add index field to support chained handlers

mostaphaRoudsari:
One of the creative usage of handlers that [came up](https://github.com/pollination/pollination-alias/blob/21a0db5569c739cfb4f0de6a1dc8b3f1721bfd4c/pollination/alias/outputs/daylight.py#L25-L38) during the development was to use handlers one after another. Luckily the current schema could support this case:

```python
    OutputAlias.linked(
        name='results',
        platform=['rhino'],
        handler=[
            # Preload results 
            IOAliasHandler(
                language='python',
                module='pollination_handlers.outputs.daylight',
                function='read_df_from_folder'
            ),
            # load preloaded outputs to Rhino with following method
            IOAliasHandler(
                language='csharp', module='Pollination.RhinoHandlers',
                function='LoadMeshBasedResultsToRhino'
            )
        ]
    )
```

The current implementation doesn't indicate which handler should be executed first. In this case since @MingboPeng was the developer of both the Rhino plugin and the Alias he knew how to get it to work but it wouldn't work otherwise.

This commit adds an optional index field to IOALiasHandler object to support supporting chained handlers. Default value will be set to 0. That means the code above will look like this:

```python

    OutputAlias.linked(
        name='results',
        platform=['rhino'],
        handler=[
            # Preload results 
            IOAliasHandler(
                language='python',
                module='pollination_handlers.outputs.daylight',
                function='read_df_from_folder', index=0
            ),
            # load preloaded outputs to Rhino with following method
            IOAliasHandler(
                language='csharp', module='Pollination.RhinoHandlers',
                function='LoadMeshBasedResultsToRhino', index=1
            )
        ]
    )
```

This change will make it explicit that the Rhino platform should use the first handler for reading the results and execute that in Python - then pass the results to the second handler and run that one in C# which is the intended behavior.

### DIFF
--- a/queenbee/io/common.py
+++ b/queenbee/io/common.py
@@ -410,3 +410,12 @@ class IOAliasHandler(BaseModel):
         description='Name of the function. The input value will be passed to this '
         'function as the first argument.'
     )
+
+    index: int = Field(
+        0,
+        description='An integer to set the index for the order of execution. This input '
+        'is only useful when there are more than one handler for the same platform and '
+        'the output of one handler should be passed to another handler. This is also '
+        'called chained handlers. By default all the handlers are indexed as 0 assuming '
+        'they are not chained.'
+    )


queenbee/io/common.py
```diff
@@ -410,3 +410,12 @@ class IOAliasHandler(BaseModel):
         description='Name of the function. The input value will be passed to this '
         'function as the first argument.'
     )
+
+    index: int = Field(
+        0,
```
AntoineDao, line 415: Any chance we can replace this a nullable field rather than a default one? This will make the recipe `digest` more reliable when comparing with digests already present on the database.

```python
index: Optional[int] = Field(None, description='An integer...'
```

AntoineDao (commented):
Hi @mostaphaRoudsari, can we slightly change this implementation to make it safer for the backend to integrate?

queenbee/io/common.py
mostaphaRoudsari: Hi @AntoineDao, Thanks for the note. I should have waited for your before merging this in. I remember we had issues with Optional when we used it in Honeybee Schema: https://github.com/ladybug-tools/honeybee-schema/issues/121#issuecomment-625302172 - optional doesn't really translate to OpenAPI and that can cause inconsistencies.

I know that this might generate a new hash for the current packages the first time but it will be fine after that.

AntoineDao: I think waiting would have been helpful here. I'm a bit nervous about what will happen when we merge this into pollination-server so I will push it now and we can keep an eye on it to see if it breaks anything

mostaphaRoudsari: Sounds good! Based on what we talked about before - adding new optional fields should be fine but you never know. 🤞

PR #366, opened 2021-09-15
fix(plugin): function cannot use . as input path 

AntoineDao:
This change is required in order for the changes implemented in pollination/queenbee-argo#135 to work. These changes were necessary to solve issues related to output artifacts being zipped, then unzipped before being pushed to an S3/GCS bucket.

I don't think we need to dive into the details of what, where and why here but essentially we can no longer set the input path of an artifact for a function to be the current working directory of the function (ie: `"path": "."` is not longer allowed for **function inputs**).

## Handling Backwards Incompatible Change
I wanted to make sure this change was palatable so I checked whether any plugins currently use this feature and only the `honeybee-radiance/split-grid-from-folder` function uses an input artifact which is saved to a path of `"."`. Furthermore, this function is not used in any of our existing recipes. 

If we merge this change as is and update the Pollination Server with it we will be introducing schema validation errors for the entire `honeybee-radiance` plugin despite the fact that only one function is incorrect. I propose that we simply delete this function from the existing plugins saved in the database and then merge this change in.

Let me know what you think!


queenbee/io/inputs/function.py
```diff
@@ -98,6 +98,12 @@ class FunctionFolderInput(DAGFolderInput):
         ' This path is relative to the working directory where the command is executed.'
     )
 
+    @validator('path')
+    def not_workspace_path(cls, v):
+        if v == '.':
+            raise ValueError('Input path for a function file cannot be "."')
```
mostaphaRoudsari, line 104:
```suggestion
            raise ValueError('Input path for a function file or folder cannot be "."')
```

tests/assets/functions/invalid/invalid_input_path.error
```diff
@@ -0,0 +1 @@
+Input path for a function file cannot be "."
```
mostaphaRoudsari, line 1:
```suggestion
Input path for a function file or folder cannot be "."
```

mostaphaRoudsari (approved):
The code looks good to me. I have a minor suggestion for the error message.

I'm fine with the suggestion to remove the function from the plugin - but we also have to remove it from the upcoming version and push it to Pollination before merging this PR.

chriswmackey (commented):
LGTM!

PR #380, opened 2022-01-06
fix(deps): bump pydantci to 1.9.0

mostaphaRoudsari:
LBT libraries are moved to 1.9.0 and without updating the dependency here we will have conflict issues for local runs.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-pydantic==1.8.2
+pydantic==1.9.0
 PyYAML==5.4.1
 jsonschema==3.2.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-pydantic==1.9.0
+pydantic>=1.8.2
 PyYAML==5.4.1
 jsonschema==3.2.0


mostaphaRoudsari:
@AntoineDao, do you foresee any possible issues here? All the tests are passing.

mostaphaRoudsari:
cc: @chriswmackey, @MingboPeng

AntoineDao:
I'm not sure what the key differences are between 1.8 and 1.9 so unclear if they introduced backwards incompatible code. If you guys haven't had any issues and tests pass here it *should* be safe. I would be more comfortable with a `pydantic>=1.8.2` instead so that other dependent libraries aren't forced to upgrade because the rest of ladybug tools has. Otherwise we're going to have to update[ `pydantic-openapi-helper`](https://github.com/ladybug-tools/pydantic-openapi-helper/blob/master/requirements.txt) and I'm not sure what misery that might entail 😛

Is that a good compromise?

AntoineDao (commented):
I'm not sure what the key differences are between 1.8 and 1.9 so unclear if they introduced backwards incompatible code. If you guys haven't had any issues and tests pass here it should be safe. I would be more comfortable with a `pydantic>=1.8.2` instead so that other dependent libraries aren't forced to upgrade because the rest of ladybug tools has. Otherwise we're going to have to update pydantic-openapi-helper and I'm not sure what misery that might entail 😛

Is that a good compromise?

mostaphaRoudsari:
That's a good compromise in my opinion and resolved the issue at hand. I made the change and will merge it in if the tests pass.

PR #574, opened 2024-05-12
ci(pypi): use PyPi token instead of username and password

AntoineDao:
Apparently username and password is no longer a supported deployment method: https://github.com/pollination/queenbee/actions/runs/9054289912/job/24873984153#step:8:3423

@mostaphaRoudsari , could I get you to add a new PyPi API Token scoped to this repository and name is `PYPI_TOKEN`? Here are the instructions to create the token: https://pypi.org/help/#apitoken

I suspect this is created and managed by the ladybugbot account. This is part of some upgrades I am making that will allow me to upgrade pollination server to python 3.12.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,8 +60,8 @@ jobs:
           echo "::set-output name=tag::$nextRelease"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
-          PYPI_USERNAME: ${{ secrets.PYPI_USERNAME }}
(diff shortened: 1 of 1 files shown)


mostaphaRoudsari:
@AntoineDao, we already had to make this change for other Pollination packages. You can use these two secrets.

image

The command line can stay the same. Here is an example:

https://github.com/pollination/annual-energy-use/blob/ff7cf9c159f4374dbd10d3f22eebbfe8a1e39772/.github/workflows/ci.yaml#L31-L38


P.S. now that you decided to change this repository to `poetry`, I wanted to make sure you are aware of `uv`: https://pypi.org/project/uv/

Streamlit started to switch to it because of the performance benefits.

https://twitter.com/ThiagoTDotCom/status/1780379729100718360

